Issue #27, opened 2015-02-17
System.getProperty fails in security-sensitive environments (like launching via java webstart)


jeffgran:
Solution is to use jruby's SafePropertyAccessor to access system properties. I'm working on a pull request.

PR #28, opened 2015-02-17
use SafePropertyAccessor to access properties instead of directly

jeffgran:
In certain contexts, java will throw an exception if you try to read an "unsafe" system property. Jruby already has a mechanism to safely try to read them and fail silently if it won't work. This just updates jruby-openssl to use that same safety mechanism.

fixes https://github.com/jruby/jruby-openssl/issues/27


jeffgran:
https://github.com/jruby/jruby-openssl/issues/27

kares:
looking good, thank you!

Issue #10, opened 2014-10-03
maybe we need some option to not load any jars


kares:
... to allow users to suply BC jars from somewhere in the CP
##

jeffgran:
:+1:

camlow325:
:+1: In Puppet Labs' Puppet Server product, we have this same need - to supply our own newer BC jar because we have other code in our app that needs a newer version than what is bundled in JRuby by default.  We're currently working around this by explicitly requiring 'jar-dependencies' and setting the JARS_NO_REQUIRE environment variable to "true" since later jopenssl versions don't require 'jar-dependencies' automatically.  We'd like to have a more direct way to specify that jars not get loaded which doesn't involve having to exercise 'jar-dependencies' directly.  What actually does the jar loading is an implementation detail that we don't really care about directly as we're mostly concerned with just turning it off.

mkristian:
jar-dependencies should be used to load the jars if the gem is available -
it comes as default gem with jruby-1.7.16+ or so. otherwise the control of
those jars depends on whether I ```require 'jar-dependencies'``` before
loading jopenssl or not.

kares:
think it makes sense even with jar-dependencies around, since jruby-openssl is kind of a must gem in all thinkable (and non-thinkable) use-cases due how the Ruby the eco-system relies on `require 'openssl'`

actually have an embed use-case of my own starting several Rubies (JRuby is custom built and gems are packed up for isolation) ... lately I found myself wanting to experiment with a custom built BC jar. think I'll go for a (unofficial) **-Djruby.openssl.load.jars=false** system property

mkristian:
jar-dependencies already has a way to pick jar versions and the next
release will have a way to tell not to load jars - i.e. decalre them
provided.

 my point here is not so much the extra system property but that the gem
ignores jar-dependencies when the gem is available (like via jruby itself).

and with 9k there will be a similar situation with snakeyaml jar which is
used for psych (via require 'yaml') so we need a common way to load or not
to load such jars. dito with jline-2.11jar from readline.

kares:
> my point here is not so much the extra system property but that the gem ignores jar-dependencies when the gem is available (like via jruby itself).

we decided previously not to load jar-dependencies explicitly since that is something the user should do, so I guess that is what you're referring to (otherwise we're using it when it's already loaded).

[...]

PR #29, opened 2015-02-18
try loading bouncycastle jars from classpath first, before loading direc...

jeffgran:
...tly

Re: Issue https://github.com/jruby/jruby-openssl/issues/10

I ran into this today. The new security requirements from Oracle are extremely restrictive. One of the  things that's not allowed any more is java.net.NetPermission("specifyStreamHandler"), which means that jruby's JRubyClassLoader, which uses a custom streamhandler called CompoundJarURLStreamHandler to load jars from inside another jar, does not work.

I was able to get around this issue by, like you mentioned, putting the BC jars on the classpath (in my JNLP file directly as lib jars) and "hacking" `jopenssl/load` to try a regular `require` before the direct `load`s that it already does.

Let me know what you think. With this fix, I was actually able (finally!) to get our jruby-based app to launch again under java 1.7u51+. I'm very willing (eager, in fact) to work with you on this to get something that you like and that will work for my use case.

### DIFF
--- a/lib/jopenssl/load.rb
+++ b/lib/jopenssl/load.rb
@@ -11,10 +11,17 @@ begin
   require_jar( 'org.bouncycastle', 'bcprov-jdk15on', version )
   bc_jars = true
 rescue LoadError
-end if defined?(Jars) && ( ! Jars.skip? ) rescue nil
+end if (defined?(Jars) && ( ! Jars.skip? ) rescue nil)
 unless bc_jars
-  load "org/bouncycastle/bcpkix-jdk15on/#{version}/bcpkix-jdk15on-#{version}.jar"
-  load "org/bouncycastle/bcprov-jdk15on/#{version}/bcprov-jdk15on-#{version}.jar"
+  begin
+    # try regular require first
+    require "bcpkix-jdk15on-#{version}.jar"
+    require "bcprov-jdk15on-#{version}.jar"
+  rescue LoadError
+    # load from here
+    load "org/bouncycastle/bcpkix-jdk15on/#{version}/bcpkix-jdk15on-#{version}.jar"
+    load "org/bouncycastle/bcprov-jdk15on/#{version}/bcprov-jdk15on-#{version}.jar"
+  end
 end
 
 require 'jruby'


kares:
Thanks Jeff, could you maybe confirm/explain how this makes a difference for you ?
as I pretty much consider `require 'with.jar'` the very same as a `load` (adds jars to JRuby's class-loader `$CLASSPATH`) except it "pollutes" `$LOADED_FEATURES` ... so you actually needed to know our internal BC .jar versions and loaded them up front or marked them as loaded in `$LOADED_FEATURES` ?

we (cc @mkristian) already tried to approach cases when BC is installed (on the class-path) - that turned out introducing more issues than solving and did not stay in ... 930ee9b5

since you mentioned this actually relates to JDK internal (security) changes, could you confirm that it worked on a previous 1.7uXX (maybe explain your use-case a bit in terms of how you're loading / class-loaders) ? as this could be affecting other JRuby pieces that rely on the loading internals ...

mkristian:
let just talk about the CompoundJarURLStreamHandler. I worked so much on classloader in jruby the last time but I NEVER there is streamhandler registered by jruby. how ignorant of me ;) - that is place I should have looked for the websphere problems in the past.

but on the other hand JRuby just works fine without using any jar: like urls with its loadService. in fact it is almost default (one PR is pending) in 9K to use only uri:classloader: which is NOT a stream-handler.

to use this new protocol with jruby-1.7.16+ you need either use ```$LOAD_PATH << 'uri:classloader:/'``` or ```ENV['JARS_HOME']='uri:classloaser:/'; require 'jar-dependencies'```. in both cases the loading of the jar will be done directly via the jrubyclassloader itself.

if you want to use BC from the classpath you can ```require 'jar-dependencies'; Jars.freeze_loading``` which means no more loading via ```require_jar``` statements. this also leads me to the question: IMO the ```require "bcpkix-jdk15on-#{version}.jar"``` (and the other require) from your PR should not be necessary since it is already on the classpath. can this be the case ?

does this security restriction ```java.net.NetPermission("specifyStreamHandler")``` throw any exceptions on the jruby side ?

mkristian:
one last thing with jruby-9k the resources from the stdlib which includes the BC jars already using the uri:classloader: protocol. and I was wrong with my suggestions:

```$LOAD_PATH.unshift 'uri:classloader://META-INF/jruby.home/lib/ruby/shared'``` or ```ENV['JARS_HOME']='uri:classloaser://META-INF/jruby.home/lib/ruby/shared';require 'jar-dependencies'```

**Comment by jeffgran** [comment]
Hi guys! Thanks so much for your thoughts. @mkristian your tip worked like magic. No more "specifyStreamHandler" issues. (now I get "accessDeclaredMembers" permission denied instead, which I don't understand, but no matter, I worked around it)

After working on this all day again today, I have it working without this change. So I'm going to close this pull request. But I thought I would hijack this thread to explain how I have it working for anyone who might come across this, and so you guys can see what I'm working with here :).

One problem that's particular to the BC jars is that they come already signed. So during my build I have to un-sign them (`zip -d bcxx.jar META-INF/*.DSA;zip -d bcxx.jar META-INF/*.SF`).

Then I have to sign ALL the jars that will be used in my application (if any jars are nested inside others, just signing the outer jar doesn't count). So nstead of going this route, the workflow below just pulls all the jars out to the top-level, and signs them there. It's easier and more explicit, and I've never actually got it fully working by re-packaging e.g. the jruby-complete.jar with all the nested jars signed.

So then I have to remove all the nested jars, e.g. `zip -d jruby-complete-xxx.jar *.jar`.

# working example

Here's my current working bootstrap code to get jruby, openssl, and all my jars and gems required while running from a jnlp:

## `myapp.jnlp`:

```XML
<jnlp codebase="http://localhost:3000/">
  <information>
  ...
  </information>
  <security>
    <all-permissions/>
  </security>
  <resources>
    <jar href="myapp.jar" main="true"/>
    <jar href="bcpkix-jdk15on-1.50.jar"/>
    <jar href="bcprov-jdk15on-1.50.jar"/>
    <jar href="jopenssl.jar"/>
    <!-- my gems built into a jar via jarjar ant task (com.tonicsystems.jarjar.Jakarta's) -->
    <jar href="gems.jar"/>
    <!-- ... etc., ALL THE JARS you will ever need. Jars cannot be nested inside each other. They must all be listed here as top-level jars resources. -->
  </resources>
  <application-desc main-class="lib.Main" />
</jnlp>
```

## bootstrap code inside `myapp.jar`:

```ruby
# these can cause us to load gems from the local system even from inside a jar in a jnlp!
ENV['GEM_PATH'] = ''
ENV['GEM_HOME'] = ''

# manually set up the load path so we're in full control of it.
$LOAD_PATH.clear

# bundler/setup was generated via `bundle install --standalone`
require 'uri:classloader://bundler/setup' # sets up gem load paths from inside gems.jar

$LOAD_PATH << 'uri:classloader://META-INF/jruby.home/lib/ruby/1.9'
$LOAD_PATH << 'uri:classloader://META-INF/jruby.home/lib/ruby/1.9/site_ruby'
$LOAD_PATH << 'uri:classloader://META-INF/jruby.home/lib/ruby/shared'

# where my code lives, inside myapp.jar
$LOAD_PATH << "uri:classloader://lib"

# this adds e.g. "http://localhost:3000" to the load path, so that the jars specified in the jnlp can get loaded from here.
$LOAD_PATH.unshift File.dirname(self.to_java.get_class().protection_domain().code_source().location().to String())

require 'jopenssl.jar'
require 'bcpkix-jdk15on-1.50.jar'
require 'bcprov-jdk15on-1.50.jar'
# etc... require ALL THE JARS

# remove the "protection domain" from the load path because it causes subsequent `require` statements to make HTTP requests, taking forever.
prev = $LOAD_PATH.clone;$LOAD_PATH.clear;prev.each { |p| ($LOAD_PATH << p) unless /^http/ === p }

require 'jar-dependencies' # this is from jruby. it just contains the ::Jars constant. it doesn't actually load any jars.
Jars.freeze_loading # tell jruby/jruby-openssl to not try loading any more jars. we already did.

require 'openssl' # now it will work! praise the programming gods!

MyApp.do_stuff
```

mkristian:
wow ! many thanks for write up.

I am surprised that a http url on $LOAD_PATH does do the http request :)

PR #32, opened 2015-02-24
add Gemfile and specify ruby-maven dependency

jeffgran:
As a mostly pure-ruby guy, it took me a while to figure out how to run the rake tasks to build the jar. This should help future developers figure out how to help contribute.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org'
+
+# Specify your gem's dependencies in the gemspec
+gemspec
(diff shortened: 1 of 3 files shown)


mkristian:
thanx - merged manually and removed the Gemfile.lock which we do not want on a library.

jeffgran:
Ah, yeah that makes sense, sorry. Maybe should add Gemfile.lock to .gitignore instead?

mkristian:
added ;)

PR #34, opened 2015-02-27
Use ByteList instead of StringBuilder. Fixes #33

### LINKED ISSUE #33: HMAC digest is incorrect when data contains invalid characters
```ruby
require 'openssl'

key = "ss"
data = "\xFF"
algo = OpenSSL::Digest::Digest.new("sha256")

digest1 = OpenSSL::HMAC.digest(algo, key, data)

hmac = OpenSSL::HMAC.new(key, algo)
hmac.update(data)
digest2 = hmac.digest

if digest1 == digest2
  p "digests are equal"
else
  p "digests are not equal"
end
```

The reason is ruby string is marshalled to java string when appending data into StringBuilder (https://github.com/jruby/jruby-openssl/blob/master/src/main/java/org/jruby/ext/openssl/HMAC.java#L182)

thedebugger:
StringBuilder append method causes rubystring conversion to java string

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/HMAC.java
+++ b/src/main/java/org/jruby/ext/openssl/HMAC.java
@@ -129,7 +129,7 @@ public class HMAC extends RubyObject {
 
     private Mac mac;
     private byte[] key;
-    private final StringBuilder data = new StringBuilder(64);
+    private ByteList data = new ByteList(64);
 
     @JRubyMethod(visibility = Visibility.PRIVATE)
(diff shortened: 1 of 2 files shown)


kares:
@thedebugger najs ... any chance we get a piece of code from #33 wrapped as a [test](https://github.com/jruby/jruby-openssl/blob/master/src/test/ruby/test_hmac.rb) ?

thedebugger:
Sure, I'll add it.

PR #37, opened 2015-04-03
Report better accept handshake errors instead of "General SSLEngine problem"

driskell:
Myself and some users of my program have been getting errors sometimes when trying to accept connections.

Within ruby, the error appears as OpenSSL::SSL::SSLError, "General SSLEngine problem".

After some lengthy diagnosis it turns out the problem is the client was sending a client certificate. On the server side, the default verify_mode is VERIFY_PEER. Thus, with no way to verify it, it cannot handshake. But rather than saying "certificate verify failed" in the accept exception, it says "General SSLEngine problem".

Looking in the java SSL and jruby-openssl code it seems the SSL stack throws in different ways and sometimes "loses" the diagnostic message (certificate verify failed). In jruby-openssl when calling connect this appears to be handled... it traces back through the exception causes to the first non-SSLHandshakeException exception, and raises a ruby exception based on that.

This patch modifies accept so it does the same as connect. This fixes my issue and in my tests my program now correctly reports "certificate verify failed".

Thanks!

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -108,6 +108,19 @@ public class SSLSocket extends RubyObject {
         return SSL.newSSLError(runtime, message);
     }
 
+    private static RaiseException newSSLErrorFromHandshake(Ruby runtime, SSLHandshakeException exception) {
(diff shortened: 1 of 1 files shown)


kares:
Excellent, thank you Jason!

PR #38, opened 2015-04-04
Add SSLSocket ssl_version property like MRI has

### LINKED ISSUE #26: NameError: uninitialized constant OpenSSL::SSL::Session
Background - 

I'm working on some security-related things and want to verify what TLS/SSL protocol has been negotiated after a connection is made before I start trusting it. The world of TLS libraries are a bit inconsistent (ruby openssl has `SSLContext#ssl_version=` to set a specific version, Go tls has `MinVersion` for a minimum version). To avoid confusion, I want to ask the socket about its session before I move forward sending sensitive data over the encrypted channel.

Situation -

To do this check, SSLSocket has a `#session` to get me the `OpenSSL::SSL::Session` object. To be  Under MRI, this seems to work, even if it's quite obscure. For example, to ask what was negotiated, `Session#to_text` will tell me.

```ruby
>> OpenSSL::SSL::SSLContext::METHODS.reject { |x| x.to_s =~ /(server|client)/ }.each { |ver|  ctx = OpenSSL::SSL::SSLContext.new; ctx.ssl_version = ver; s = OpenSSL::SSL::SSLSocket.new(TCPSocket.new("google.com", 443), ctx); s.connect; p ver => s.session.to_text.split("\n").grep(/Proto/).first.split(/ *: */).last }
{:TLSv1=>"TLSv1"}
{:TLSv1_2=>"TLSv1.2"}
{:TLSv1_1=>"TLSv1.1"}
{:SSLv3=>"SSLv3"}
{:SSLv23=>"TLSv1.2"}
```

This fails under JRuby 1.7.17 during the `OpenSSL::SSL::SSLSocket#session` call because OpenSSL::SSL::Session is not defined.

---

This is NOT a blocker. While I don't yet know of a run-time work-around, I can work around this by doing more extensive testing which I'll need to do anyway.

driskell:
When using Ruby 2.0 we can call .ssl_version on an SSLSocket to see what protocol was negotiated in the handshake. With JRuby this is not possible.

This adds the ssl_version method so we can now do the same thing.

I can't say for certain, but this might be what @jordansissel needs in order to resolve #26 too.

(The Ruby commit: https://github.com/ruby/ruby/commit/060184c347822b11dff3db6bef915c04a564c4e4)

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -854,6 +854,12 @@ public class SSLSocket extends RubyObject {
         return getRuntime().getNil(); // throw new UnsupportedOperationException();
     }
 
+    @JRubyMethod
+    public IRubyObject ssl_version() {
+        if ( engine == null ) return getRuntime().getNil();
+        return getRuntime().newString( engine.getSession().getProtocol() );
+    }
+
     private SocketChannel getSocketChannel() {
         return (SocketChannel) io.getChannel();
     }
--- a/src/test/ruby/ssl/test_ssl.rb
+++ b/src/test/ruby/ssl/test_ssl.rb
@@ -83,6 +83,10 @@ class TestSSL < TestCase
   end
 
   def test_ssl_version_sslv3
+    skip('Disable SSLv3 test in CI as it currently fails on some JVM versions') unless ENV['CI'].nil?
+    # This test appears to fail on Oracle JDK 1.7.0_76 but not Oracle JDK 1.6.0_65
+    # The test (client) reports Connection reset by peer
+    # The server reports "No appropriate protocol (protocol is disabled or cipher suites are inappropriate)"
     ctx_proc = Proc.new do |ctx|
       ctx.ssl_version = "SSLv3"
     end
--- a/src/test/ruby/ssl/test_ssl.rb
+++ b/src/test/ruby/ssl/test_ssl.rb
@@ -82,4 +82,30 @@ class TestSSL < TestCase
     end
   end
 
+  def test_ssl_version_sslv3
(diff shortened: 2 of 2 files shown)


kares:
Thanks again Jason, the removed test should be valuable (for some time period) locally (until we hit the same with a JDK update) ... could you maybe add a `unless ENV['CI']` or something - no biggie we can do it on top of the commits as they are. related to #26 we do not emulate `OpenSSL::SSL::Session` (yet)

driskell:
I've had it call skip() if running in Travis.

Tested with 1.6.0_65 and 1.7.0_76 Oracle JDK and it seems when running jruby under 1.6.0_65 JDK it works. When running under 1.7.0_76 JDK the SSL server reports "No appropriate protocol (protocol is disable or cipher suites are inappropriate)" - got that error by adding diagnostic in test_helper's server_loop.

driskell:
Looks like since 1.7.0_75, SSLv3 is disabled by default. So there's the culprit. Though I don't know why jruby-openssl isn't able to re-enable it internally...
http://www.oracle.com/technetwork/java/javase/7u75-relnotes-2389086.html

driskell:
Though I guess the ultimate problem is enabling SSLv3 again would enable it for other parts of the JVM, and is thus too dangerous. So something to live with I guess! Maybe the SSLv3 test should just be removed - it was really only to test the return value of ssl_version.

kares:
first commit is in https://github.com/jruby/jruby-openssl/commit/6886484dfb8a5954eb00e8854da8f2bb1e0d7d81 ... second one was kept out in favour of https://github.com/jruby/jruby-openssl/commit/cfeba095e8f7823dc040ec2b1f517385a74296a2 ... thanks!

driskell:
Thanks! :+1:

PR #41, opened 2015-04-30
Mimic MRI's openssl when jopenssl require fails.

voxik:
Some libraries (such as RubyGems [1]) rescues openssl load errors and
adjust their behavior depending if the openssl is properly required.
However, prior this commit, require 'openssl' succeeded just fine, but
the subsequent requires failed, which makes RubyGems (and similar) to
fail unexpectedly.

Please note that I submitted similar jruby/jruby/pull/2894 for 1.7 branch.

[1] https://github.com/rubygems/rubygems/blob/master/lib/rubygems/security.rb#L14

### DIFF
--- a/lib/openssl.rb
+++ b/lib/openssl.rb
@@ -1,1 +1,5 @@
-require 'jopenssl/load'
+begin
+  require 'jopenssl/load'
+rescue LoadError => e
+  e.message.sub! 'jopenssl/load', 'openssl'
+end


voxik:
Actually, RubyGems checks for ```e.path == 'openssl'``` [1] as well, so this is probably not complete patch :/ Not sure if there is better way how to make Ruby think that 'openssl' can't be required.


[1] https://github.com/rubygems/rubygems/blob/master/lib/rubygems/security.rb#L13

kares:
thanks @voxik, have you tried the other way ... submitting a patch to RGs itself explaining the issue?

voxik:
@kares No, I have not, since I don't believe that RubyGems are the only library in the world using Ruby's OpenSSL. There are probably others which are doing similar check. And to be honest, I believe that most of the SW is tested against MRI and this is MRI compatibility issue.

voxik:
Actually, since output of jruby/jruby#2894 is totally different then anticipated, this might become less critical.

kares:
@voxik disagree ... most code does `defined? OpenSSL` or `require 'openssl'` rescue-ing a `LoadError`. feel free to prove me wrong with another example, otherwise RG folks might be happy to fix it - if there's a real issue. you should maybe present us a valid use-case where this was an issue under JRuby.

voxik:
You describe almost precisely what RubyGems does. But if you check the links I provide above, you'd notice that they in addition check what actually was not loaded.

kares:
@voxik did check I was asking for another example this being done in the wild (if you read my comment), until than we shall assume only RGs does the "hack" and try to come up with a solution on their side ...

voxik:
@kares Thanks for opening the RubyGems ticket. Hopefully they will answer soon :)

deivid-rodriguez:
I assume this is no longer an issue since https://github.com/rubygems/rubygems/commit/babf9431440fbb8cbb6391a529f3d44bf355f023 and/or https://github.com/rubygems/rubygems/commit/43a279dab380fb789b6bb46e77682e90d152d866?

PR #54, opened 2015-07-08
fixing oaep encryption to use correct algorithm

balmas:
Proposed fix for problem reported in https://github.com/jruby/jruby/issues/1819

I have to confess that writing a reasonable java test for this change is beyond me, but at [line 99](https://github.com/balmas/jruby-openssl/blob/f3a0964ecad5b788c82431b3a4863957d9f1b233/src/main/java/org/jruby/ext/openssl/PKeyRSA.java#L99) the constant `PKCS1_OAEP_PADDING` is  set to 4. The test on the changed line was previously selecting the algorithm `/ECB/OAEPWithMD5AndMGF1Padding` for oaep, but this does not work against data that declares itself as encoded according to http://www.w3.org/2001/04/xmlenc#rsa-oaep-mgf1p and my reading of section 5.4.2 of the [xmlenc spec](http://www.w3.org/TR/2002/REC-xmlenc-core-20021210/Overview.html) says that this must always be used with SHA1 (and not MD5).

If desired, I can provide sample data and ruby code that demonstrates the problem and verifies the fix.


kares:
looking good, probably an old compatibility issue ... possibly we would need to make sure it does not regress on 1.8.7 since JRuby 1.7.x still supports --1.8 mode.

balmas:
Hi I added the test to confirm creation of a Cipher object with RSA/ECB/OAEPWithSHA1AndMGF1Padding and it doesn't fail on regressions with 1.8 mode (e.g. mvn test -Djruby.versions=1.7.12 -Djruby.modes=1.8 -Dbc.versions=1.47) although I'm not sure that's really a meaningful test.

Will add sample data and ruby code demonstrating the problem shortly.

balmas:
Okay, ruby code and test data added (in src/test/ruby/oaep).

Confirmed passes with new code, and raises error with old.  No errors reported running test in 1.8 mode.

Let me know if I can do anything else. Thanks!

PR #60, opened 2015-08-06
Test little things

mkristian:


mkristian:
@kares please have a look at those "little" changes. this SecurityHelper from here is the one which get used by https://github.com/lookout/fast-rsa-engine/blob/master/src/main/java/org/jruby/ext/openssl/SecurityHelper.java

so all those changes would help to inject ciphers to be used by jruby-openssl, i.e. fast-rsa-engine would not need to load certain java classes first (kind of monkey patching the java way)

the extra reflection might be even useful for general cases (oracle-jdk)

kares:
honestly I wished to be removing parts of `SecurityHelper` not add more :) as it was introduced as an ugly hack - the motivation was to allow to roll without the crazy 'thing' of self-installing BC as a provider.

must say I was little eager for the merge - now I see room for the fast-rsa gem to exist on its own (at least for a while) ... if you really need the changes go for it, but please try not to push them too far. these public methods on "helpers" (we do not use directly) need at least some javadoc so they're not removed easily: https://github.com/jruby/jruby-openssl/commit/55761590ca5d4199fb6f3f64f1fd250568e2f0f9 ... also if the fake classes had a note and maybe a license on top that would be great. thanks!

PR #61, opened 2015-08-10
keep in line with MRI if possible

### LINKED ISSUE #49: certs + private keys location is not in line with MRI
```
/usr/lib/ssl$ ls -al
certs -> /etc/ssl/certs
misc
openssl.cnf -> /etc/ssl/openssl.cnf
private -> /etc/ssl/private
```
MRI seems to inherits this from openssl

currently things do follow this partially:
* for private keys it looks at /usr/lib/ssl/private
* cacerts it looks in $JAVA_HOME/lib/security/cacerts but not in /usr/lib/ssl/certs

ubuntu does link the /usr/lib/ssl/ files to /etc/ssl

so altogether jruby-openssl do use the openssl directories and just fall back on  $JAVA_HOME/lib/security/cacerts if there is a not other cacerts.

mkristian:
keep the default x509 certs and directories in line with MRI, only if
they do not exists fallback on cacerts from the java.home/lib/security/cacerts

fixes #49 and keeps the idea of https://github.com/jruby/jruby-openssl/commit/b914091011bc68d313f6f25a623906057c888874

Sponsored by Lookout Inc.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/x509store/X509Utils.java
+++ b/src/main/java/org/jruby/ext/openssl/x509store/X509Utils.java
@@ -28,6 +28,7 @@
 package org.jruby.ext.openssl.x509store;
 
 
+import java.io.File;
 import java.io.IOException;
 import java.math.BigInteger;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


mkristian:
@kares @headius please have a look this.

kares:
makes sense - hopefully there won't be any drawbacks if there are any we can make a system-prop switch

kares:
on a second note (without looking) will we actually use `X509_CERT_AREA` if it points to OpenSSL's cert dir since the SSL engine will have those available from Java for sure (as reported in https://github.com/jruby/jruby/issues/1953). 

... cert lookup will honor these, although by default does nothing, right?

mkristian:
```X509_CERT_AREA``` is not used but the ```X509_CERT_DIR``` and ```X509_CERT_FILE```

actually b914091011bc68d313f6f25a623906057c888874 made those two to be used.

still https://github.com/jruby/jruby/issues/1953 says paths are different then on ubuntu. will look into this.

mkristian:
did not expect to step in such a jungle: https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/

Issue #4, opened 2014-07-31
Bouncycastle 1.51 causes load error: jopenssl/load -- java.lang.VerifyError: Bad type on operand stack 


cheister:
The 1.51 Bouncy Castle release http://www.bouncycastle.org/releasenotes.html causes a load error.  
From the release notes they claim "Some changes have affected the return values of some methods" which I suspect is to blame
The error is:

```
[WARNING] ERROR:  Loading command: install (LoadError)
[WARNING]   load error: jopenssl/load -- java.lang.VerifyError: Bad type on operand stack
[WARNING] Exception Details:
[WARNING]   Location:
[WARNING]     org/jruby/ext/openssl/X509Cert.sign(Lorg/jruby/runtime/ThreadContext;Lorg/jruby/runtime/builtin/IRubyObject;Lorg/jruby/runtime/builtin/IRubyObject;)Lorg/jruby/runtime/builtin/IRubyObject; @140: invokevirtual
[WARNING]   Reason:
[WARNING]     Type 'org/bouncycastle/asn1/ASN1ObjectIdentifier' (current frame, stack[1]) is not assignable to 'org/bouncycastle/asn1/DERObjectIdentifier'
[WARNING]   Current Frame:
[WARNING]     bci: @140
[WARNING]     flags: { }
[WARNING]     locals: { 'org/jruby/ext/openssl/X509Cert', 'org/jruby/runtime/ThreadContext', 'org/jruby/runtime/builtin/IRubyObject', 'org/jruby/runtime/builtin/IRubyObject', 'org/jruby/Ruby', 'java/lang/String', 'java/lang/String', 'java/lang/String', 'java/util/Iterator', 'org/jruby/ext/openssl/X509Extensions$Extension', '[B' }
[WARNING]     stack: { 'org/bouncycastle/x509/X509V3CertificateGenerator', 'org/bouncycastle/asn1/ASN1ObjectIdentifier', integer, '[B' }
```

trejkaz:
Isn't one potential resolution for this sort of issue to prefix BC's package names so that they don't clash with a different version? :/

PR #62, opened 2015-08-14
fix load error: jopenssl/load -- java.lang.VerifyError: Bad type on operand stack when using bouncycastle 1.51 or 1.52

cheister:
This should fix https://github.com/jruby/jruby-openssl/issues/4

In BouncyCastle 1.51 they [changed the hierarchy of ASN1ObjectIdentifier](https://github.com/bcgit/bc-java/commit/737bedf272b4fb34c395f07addb72b28d4f90a13) so it is a parent of DERObjectIdentifier instead of a descendant of it.  In the code, if we pass a string id instead of the  ASN1ObjectIdentifier object then BouncyCastle will create the correct ASN1ObjectIdentifier object.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/X509Cert.java
+++ b/src/main/java/org/jruby/ext/openssl/X509Cert.java
@@ -555,7 +555,7 @@ public class X509Cert extends RubyObject {
         for ( X509Extension ext : uniqueExtensions() ) {
             try {
                 final byte[] bytes = ext.getRealValueEncoded();
-                builder.addExtension(ext.getRealObjectID(), ext.isRealCritical(), bytes);
+                builder.addExtension(ext.getRealObjectID().getId(), ext.isRealCritical(), bytes);
             }
             catch (IOException ioe) {
                 throw runtime.newIOErrorFromException(ioe);


kares:
thanks we'd appreciate a test-case esp. as we might still need to support at least BC 1.50 (> 1.50 was not fully really reviewed yet - not sure what's missing hopefully not match) and this might break older versions

cheister:
You can reproduce the error if you run `mvn verify -P test-1.7.19 -Dbc.versions=1.51` or `mvn verify -P test-1.7.19 -Dbc.versions=1.52`.

It would be very surprising if this broke previous versions since the X509V3CertificateGenerator.addExtension(String, boolean, ASN1Encodable) method is just delegating to the X509V3CertificateGenerator.addExtension(DERObjectIdentifier, boolean, ASN1Encodable) method in 1.50 and the X509V3CertificateGenerator.addExtension(ASN1ObjectIdentifier, boolean, ASN1Encodable) in 1.51 and 1.52.  https://github.com/bcgit/bc-java/blob/master/prov/src/main/java/org/bouncycastle/x509/X509V3CertificateGenerator.java#L225

kares:
understood, thanks for explaining this further.

PR #63, opened 2015-08-21
Support negotiating up to TLS1_1 and TLS1_2 when the server supports these ssl_versions

cheister:
This fixes an issue we were having where JRuby apps couldn't connect to a server running only with TLS1_1 and TLS1_2 enabled unless you specifically told the JRuby client to use TLS1_1 or TLS1_2.  The JRuby client would not negotiate up to the highest ssl version allowed by the server, it would always stop at TLS1.


cheister:
Looks like TLS1_1 and TLS1_2 are [not supported by jdk6](https://blogs.oracle.com/java-platform-group/entry/diagnosing_tls_ssl_and_https) so I added conditional logic around the tests and SSL constant to only run in java 7+

kares:
picked on master (with some updated test-excludes), thanks!

cheister:
Thanks for fixing the tests!

PR #66, opened 2015-08-24
return the same ssl_version format that is presented in OpenSSL::SSL::SSLContext::METHODS and also matches MRI

cheister:
According to [the comment](https://github.com/jruby/jruby-openssl/blob/master/src/main/java/org/jruby/ext/openssl/SSLContext.java#L187) for the OpenSSL::SSL::SSLContext::METHODS implementation we don't want to expose TLSv1.1 or TLSv1.2 as ssl_version options.  This change should keep those strings from coming out via the ssl_version accessor.  

NOTE: This change is backwards incompatible if anyone is currently checking the ssl_version against TLSv1.1 or TLSv1.2.


kares:
the test changes are NOT compatible with MRI ... e.g. on 2.2.2 I'm getting `ssl_version` as **"TLSv1.1"**

PR #71, opened 2015-09-11
More default locations

mkristian:


mkristian:
@kares @headius @enebo 

the second trail this time with tests. though it is not possible to test those default locations but it is now possible to test SSL_CERT_FILE with PEM or java keystore files. fixed a few other issues which showed up on the way.

I did not change this Lookup.java class which mixed C-style and java style error handling.

mkristian:
@kares there is probably no file extension matching with openssl since there is no need for it. it just process it. but process is tricky since it looks for the begin- and end-marker, i.e. when not found then the file is considered processed.

the only jruby-openssl could achieve the same is to first try it as java keystore and in case of an io-error process it as "pem"

currently it will produce an error when the extension does not match.

let me know if I shall add the extra fallback on a failed  keystore loading.

kares:
@mkristian does not seem to matter, hopefully no users will run into issues with the new logic and we won't need to revert

PR #75, opened 2015-10-13
Adding support for auth_tag for GCM based crypto [WIP]

trashhalo:
This pr is far from mergable but it gives us enough to keep going with our migration from mri to jruby.

The issue is java 8 does not support using an IvParameterSpec with GCM as can be seen in this bug. [SANTUARIO-392](https://issues.apache.org/jira/browse/SANTUARIO-392) so we added a conditional block to use the gcm spec object. Additionally, [ruby 2 cipher supports getting ahold of the auth_tag](http://ruby-doc.org/stdlib-2.0.0/libdoc/openssl/rdoc/OpenSSL/Cipher.html#method-i-auth_tag) but only after you call final, which we needed for our use case.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/Cipher.java
+++ b/src/main/java/org/jruby/ext/openssl/Cipher.java
@@ -46,6 +46,7 @@ import static javax.crypto.Cipher.DECRYPT_MODE;
 import static javax.crypto.Cipher.ENCRYPT_MODE;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.RC2ParameterSpec;
 
 import org.jruby.Ruby;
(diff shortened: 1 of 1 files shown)


kares:
thanks, seems good. I believe a passing test or two could make this merge material ...

trashhalo:
@kares Sure I can clean this up!

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project jruby-openssl: Compilation failure: Compilation failure:
[ERROR] /home/travis/build/jruby/jruby-openssl/src/main/java/org/jruby/ext/openssl/Cipher.java:[49,25] cannot find symbol
[ERROR] symbol  : class GCMParameterSpec
[ERROR] location: package javax.crypto.spec
[ERROR] /home/travis/build/jruby/jruby-openssl/src/main/java/org/jruby/ext/openssl/Cipher.java:[49,25] cannot find symbol
[ERROR] symbol  : class GCMParameterSpec
[ERROR] location: package javax.crypto.spec
[ERROR] /home/travis/build/jruby/jruby-openssl/src/main/java/org/jruby/ext/openssl/Cipher.java:[1048,29] cannot find symbol
[ERROR] symbol  : class GCMParameterSpec
[ERROR] location: class org.jruby.ext.openssl.Cipher
```
So GCMParameterSpec was only added in java 7. Do you want me to check for its presence and use if available otherwise keep using Iv?

kares:
you can try to make this work in Java 7+ only and we're worry about making it Java 6 compatible later ...

trashhalo:
@karesIf I am reading the travis output right build passed on all openjdk7+

trashhalo:
Closing the pr till we can get an end to end test working (doesnt seem to be working atm) jruby encrypted mri decrypted

PR #82, opened 2016-03-08
Instantiate both the RSA private and public keys when setting parameters

philr:
The RSA `e` (exponent) and `n` (modulus) parameters are initially backed by the `rsa_e` and `rsa_n` fields. These fields are then used to instantiate the internal `privateKey` and `publicKey` objects in `generatePrivateKeyIfParams()` and `generatePublicKeyIfParams()`. However, both `rsa_e` and `rsa_n` are set to `null` after use. This means that once either of the private or public keys has been created, the other cannot be. Consequently, the RSA object can only be used for private or public functions, but not for both.

For example:

```ruby
orig = OpenSSL::PKey::RSA.generate(2048)
copy = OpenSSL::PKey::RSA.new
[:e, :n, :d, :p, :q, :iqmp, :dmp1, :dmq1].each do |param|
  copy.send("#{param}=", orig.send(param))
end

# e and n were set first, copy will now act only as a public key

# public_encrypt will succeed
copy.public_encrypt('Test')

# private_encrypt will fail with an 'OpenSSL::PKey::RSAError: private key needed' exception
copy.private_encrypt('Test')
```

When the above code is run using the MRI standard library, both the `public_encrypt` and `private_encrypt` calls succeed.

This pull request modifies `generatePrivateKeyIfParams()` and `generatePublicKeyIfParams()` so that they will obtain the `e` and `n` parameters from either `privateKey` or `publicKey` if `rsa_e` and `rsa_n` have already been consumed.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/PKeyRSA.java
+++ b/src/main/java/org/jruby/ext/openssl/PKeyRSA.java
@@ -648,17 +648,19 @@ public class PKeyRSA extends PKey {
         return getRuntime().getNil();
     }
 
+    private BigInteger getPublicExponent() {
+        if (publicKey != null) {
+            return publicKey.getPublicExponent();
+        } else if (privateKey != null) {
+            return privateKey.getPublicExponent();
+        } else {
+            return rsa_e;
+        }
+    }
+
     @JRubyMethod(name="e")
     public synchronized IRubyObject get_e() {
-        RSAPublicKey key;
-        BigInteger e;
-        if ((key = publicKey) != null) {
-            e = key.getPublicExponent();
-        } else if(privateKey != null) {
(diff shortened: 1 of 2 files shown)


philr:
The build failure seems to be unrelated to this change (the same job was failing with the same error before).

kares:
Thanks Phil, this is looking good. we probably do not need those private methods to be `synchronized`.

philr:
@kares I have amended my commit to remove `synchronized` from the `getPublicExponent()` and `getModulus()` private methods.

Travis CI has done a new build. However, this time the jobs have all failed with an unrelated issue initializing the build environment.

kares:
yep, restarted the build and all seems well, thank you for your effort.

PR #83, opened 2016-03-08
Allow DSA private keys to be initialized from parameters

philr:
It is currently not possible to initialize an instance of `OpenSSL::PKey::DSA` for use as a private key by setting the individual parameters. This is because the `priv_key=` method is missing and `setKeySpecComponent()` only handles the creation of the `publicKey`.

This pull request adds a `priv_key=` method and modifies `setKeySpecComponent()` method to initialize `privateKey` in addition to `publicKey`.

Similarly to #82 the DSA `p`, `q` and `g` parameters are shared between the private and public keys. I've changed `setKeySpecComponent()` to read these parameters from `privateKey` or `publicKey` if necessary.


philr:
The Travis CI build has failed due to my use of the `refute` method in two of the test cases (which appears to not exist in older versions of JRuby). I will update the PR to fix the error.

philr:
I have amended my commit to replace the problematic `refute` calls. The remaining build failure is unrelated to this change.

kares:
thanks, esp. that you got us some new tests - much appreciated!

Issue #93, opened 2016-05-29
Errno::EAGAIN thrown by SSLSocket#connect_nonblock


mohamedhafez:
A few times a day I'm getting `Errno::EAGAIN` thrown [here in Net::HTTP#connect](https://github.com/jruby/jruby/blob/master/lib/ruby/stdlib/net/http.rb#L938), I'm not sure if this is a fault of the ruby code not handling `Errno::EAGAIN`, or if MRI ruby's `SSLSocket#connect_nonblock` handles `Errno::EAGAIN` internally.

I'm running a patch that catches the `Errno::EAGAIN`, sleeps 0.01, and then retries, and it works just fine, if the case is that the ruby code should really be handling this then I can submit a fix to ruby and we can close this issue. If that's not the case though and this should be handled internally by jruby's SSLSocket implementation to match MRI's behavior, then the stack trace points to this error being thrown at:

```ruby
org/jruby/ext/openssl/SSLSocket.java:225:in `connect_nonblock'
```

(I have no idea how to reproduce `Errno::EAGAIN` to see if MRI handles it or not, sorry!)

kares:
Thanks Mohamed, unfortunately I'm not sure someone would run into trying to fix this without reproducing.

mohamedhafez:
Now that I think about it, how would `SSLSocket#connect_nonblock` handle `Errno::EAGAIN` with out blocking anyway? That error literally means "wait for a bit and then retry", and a nonblocking method shouldn't be sleeping and then retrying... I think the correct fix should be dealing with the error in Net::HTTP#connect. I'm going to ask on the main ruby mailing list real quick and then probably close this issue

mohamedhafez:
Jeez ignore that last one, I obviously know nothing about nonblocking io coding:/ After reading a bunch and looking at the code, I'm guessing that if a socket would block, it should always either be wait readable or wait writable, and there shouldn't ever be a generic "this socket would block on something but I can't tell you if its reading or writing", correct? I'm trying to go through the jruby code myself and see if I can't spot where those `EAGAIN`s might happen, since I also have another couple errors I'm encountering once every few days in production that I don't know how to reproduce, and I'm hoping maybe there is a chance I could figure it out and fix them.

mohamedhafez:
[...]

PR #95, opened 2016-06-04
waitSelect improvements

mohamedhafez:
- makes `waitSelect` a bit more efficient for nonblocking calls, by pulling that code outside of RubyThread.BlockingTask 

- As reported in https://github.com/jruby/jruby-openssl/issues/93, there is a JRuby bug where there may be old selectors returned by `selectedKeys()`, this fix checks all selected keys just in case our key is in there but just isn't the first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,6 @@ pkg
 target
 build.log
 .idea
+*.iml
 *.lock
 lib/org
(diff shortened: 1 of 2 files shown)


kares:
looking good, thank you. 
right, we do not need the blocking task - would like to wait for feedback for the "leaking" selector pool issue at JRuby's. as I'd like to possibly confirm your thoughts on why its actually happening.

mohamedhafez:
@kares https://github.com/jruby/jruby/pull/3952 to fix the selector pool issue just got merged

kares:
@mohamedhafez great, thanks again!

Issue #113, opened 2016-11-21
ClassCastException error in X509Store.verify


cprice404:
I've been debugging an issue that surfaced when I filed [a PR to upgrade JRuby 1.7 to jruby-openssl 0.9.18](https://github.com/jruby/jruby/pull/4307).

The issue is that when the [jruby external tests for x509store](https://github.com/jruby/jruby/blob/jruby-1_7/test/externals/ruby1.9/openssl/test_x509store.rb#L161) are run, there is a stack trace that looks like this:

```
test_verify(OpenSSL::TestX509Store):
Java::JavaLang::ClassCastException: org.bouncycastle.jce.provider.X509CertificateObject cannot be cast to org.jruby.ext.openssl.X509Cert
    org.jruby.ext.openssl.X509StoreContext.initialize(X509StoreContext.java:132)
    org.jruby.ext.openssl.X509StoreContext.newStoreContext(X509StoreContext.java:102)
    org.jruby.ext.openssl.X509Store.verify(X509Store.java:225)
    org.jruby.ext.openssl.X509Store$INVOKER$i$0$0$verify.call(X509Store$INVOKER$i$0$0$verify.gen)
    org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:218)
```

After some investigation I discovered that there is a test that looks very similar included directly in the `jruby-openssl` repo:

https://github.com/jruby/jruby-openssl/blob/master/src/test/ossl/1.9/test_x509store.rb#L36

If I check out the 0.9.18 tag of `jruby-openssl`, and run this test locally by setting up JRuby 1.7.26 as my local jruby and running:

```
jruby -Ilib:. src/test/ossl/1.9/test_x509store.rb
```

I get the same error.  Examining the `travis.yml` file, and doing some local experimenting with the various rake tasks for running tests, it *appears* that this test is never being run by CI.

I can debug the failure itself and with any luck maybe file a PR to fix it, but I don't know how to go about modifying the code so that the test would be run in CI.  It seems like, if these tests are useful, they should be run in CI?

I'm also trying to figure out why this hasn't shown up in JRuby9k, which appears to have already been upgraded to `jruby-openssl` 0.9.18.  If I clone a fresh copy of the jruby repo and run from the HEAD of the master branch:

```
$ mvn -P bootstrap
...
$ export PATH=./bin:$PATH
$ jruby --version
[...]

PR #114, opened 2016-11-23
Fix ClassCastException in X509Store.verify

cprice404:
This commit changes the logic of the `X509StoreContext.initialize`
method, to address some test failures that were showing up in
`test_x509store.rb`.  The failures would occur during calls to
the `verify` method of `X509Store`.

It seems that the failures were introduced by the addition of the
`X509Cert.toJava` method in ecc1a05f458bdc05d6d0f6fd025f8788ded0f120 .
The `toJava` method seems to be implicitly called when iterating over
a RubyArray in Java code, and the implementation of that method
now returns a BC object rather than the expected JRuby `X509Cert`
object.

This commit changes the iteration to use the `.eltOk` method to access
the underlying objects w/o the implicit call to `toJava`.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/X509StoreContext.java
+++ b/src/main/java/org/jruby/ext/openssl/X509StoreContext.java
@@ -127,9 +127,17 @@ public class X509StoreContext extends RubyObject {
         final List<X509AuxCertificate> _chain;
         if ( ! chain.isNil() ) {
             @SuppressWarnings("unchecked")
-            final List<X509Cert> certs = (List<X509Cert>) chain; // RubyArray
+            final RubyArray certs = (RubyArray) chain;
             _chain = new ArrayList<X509AuxCertificate>( certs.size() );
-            for ( X509Cert x : certs ) _chain.add( x.getAuxCert() );
+
+            for (int i = 0; i < certs.size(); i++) {
+                // NOTE: if we use the normal java syntax for iterating over this
+                // RubyArray, the `toJava` method of the X509Cert class will be
+                // implicitly called, and that will return the BC certificate object
+                // rather than the JRuby one.
+                X509Cert c = (X509Cert) certs.eltOk(i);
+                _chain.add(c.getAuxCert());
+            }
         }
         else {
             _chain = new ArrayList<X509AuxCertificate>(4);


cprice404:
This is an attempt to address the issues described in #113 .

@kares I'm not sure if this is the right approach, and I don't fully understand how `.eltOk` works, but I was discussing the bug with @enebo in IRC and he suggested that I might need to call that method in order to avoid the implicit call to `.toJava`.

With this change, I can run the entire MRI `test_x509store.rb` suite locally and all tests pass.

I think that, if this change is accepted, it would also be a good idea to fix the CI for `jruby-openssl` and the master/9k branches of `jruby` to make sure that they are running the MRI openssl tests - I assume that they are expected to pass, since they are still run as part of the CI for JRuby 1.7.x.

cprice404:
If this doesn't seem like the best approach, and if someone who is more familiar than I am with this code isn't going to have time to look into this soon, I might be able to find a little more time to work on it next week if someone can give me some hints on a better implementation approach.  Given that the `X509Cert.toJava` method may be implicitly called during seemingly benign operations like list iteration, it seems like it might be better to avoid having it return a BC object type by default, but I don't have a very good grasp of what other considerations led to that implementation of `.toJava` to begin with so I was afraid of breaking something else if I tried to make changes there.

kares:
This is excellent Chris, thank you! I mostly introduced the `toJava` for easier debugging (e.g. while testing).
Your PR is just fine -> where in a Ruby method unlikely to get any other Java collection type thus casting to a `RubyArray` and making sure we extract the raw ruby values seems like the very best approach here.

I haven't looked at the MRI test that identified this bug (somehow assumed its not that crucial since 9K passed fine) - would you consider copy-ing the test to our end to make sure we have a regression test?

cprice404:
The test that failed is, I believe, directly copied from the MRI test suite, and it already exists in both the `jruby-openssl` source and the jruby source (both 1.7 and master branches):

https://github.com/jruby/jruby-openssl/blob/master/src/test/ossl/1.9/test_x509store.rb#L36
https://github.com/jruby/jruby/blob/jruby-1_7/test/externals/ruby1.9/openssl/test_x509store.rb
https://github.com/jruby/jruby/blob/master/test/mri/openssl/test_x509store.rb

But for whatever reason, this test *is* being run in CI for JRuby 1.7, and is *not* being run in CI for `jruby-openssl` or JRuby 9k.  I've run it in 9k and it fails.

You can run it in the `jruby-openssl` source tree via the following command, but I don't know how to add that to the CI configuration:

    jruby -Ilib:. src/test/ossl/1.9/test_x509store.rb

kares:
unfortunately the **src/test/ossl** is not run - never finished the setup (needs working excludes) and we do rely on running the full JRuby suite before the final release push.

anyways, I can handle things from here just fine. thanks for the effort!

kares:
on master as https://github.com/jruby/jruby-openssl/commit/b5bec3e62ba50eda10eeae2931f1450a8d2405b3

Issue #120, opened 2017-01-13
renegotiation_cb attribute missing on OpenSSL::SSL::SSLContext


lampad:
I'm not sure how well Java supports renegotiation under the hood. I've done quite a bit of digging, and there are very few documents (if any, certainly nothing official) on how to trigger secure renegotiation from a java client or handle it as a server that complies with the renegotiation [RFC](https://tools.ietf.org/html/rfc5746#section-3). Most of the searches I did returned panicked articles about CVE-2009-3555.

In any case, [CRuby supports setting a renegotiation callback](http://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/SSL/SSLContext.html#attribute-i-renegotiation_cb-label-Disable+client+renegotiation) that gets triggered when SSL connections are established (similar to the other callbacks on the SSLContext):

```
2.4.0 :001 > require 'openssl'
 => true
2.4.0 :002 > ctx = OpenSSL::SSL::SSLContext.new
 => #<OpenSSL::SSL::SSLContext:0x007f9e94032e30>
2.4.0 :003 > ctx.renegotiation_cb = Proc.new { "pop" }
 => #<Proc:0x007f9e9403b670@(irb):3>
```

JRuby does not:
```
jruby-9.1.7.0 :001 > require 'openssl'
 => true
jruby-9.1.7.0 :002 > ctx = OpenSSL::SSL::SSLContext.new
 => #<OpenSSL::SSL::SSLContext:0x7966baa7>
jruby-9.1.7.0 :003 > ctx.renegotiation_cb = Proc.new { "pop" }
NoMethodError: undefined method `renegotiation_cb=' for #<OpenSSL::SSL::SSLContext:0x7966baa7>
	from (irb):3:in `<eval>'
	from org/jruby/RubyKernel.java:1000:in `eval'
	from org/jruby/RubyKernel.java:1298:in `loop'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from org/jruby/RubyKernel.java:1120:in `catch'
	from /Users/dlampa/.rvm/rubies/jruby-9.1.7.0/bin/irb:13:in `<main>'
```

Is there a technical reason this attribute wasn't included in jruby-openssl? I'm willing to take a crack at implementing it, but wanted to make sure whether or not there was a decision to explicitly not include it.

---
[...]

PR #121, opened 2017-01-18
Add renegotiation cb

lampad:
An attempt to resolve: https://github.com/jruby/jruby-openssl/issues/120

Let me know if you think the callback call is in the correct place. Based on comparing what CRuby does, which basically delegates to OpenSSL, I think the current location makes the most sense, but I also played around with putting it at the start of the doHandshake() method.

I had some difficulty getting my environment set up correctly, and wasn't able to run the full ruby test suite, but the java tests and the tests in the test_ssl.rb pass. I'm hoping the CI keeps me honest.


kares:
@lampad well done, thanks!

Issue #24, opened 2014-12-29
Missing openssl support OpenSSL::OCSP


headius:
From http://jira.codehaus.org/browse/JRUBY-7197

I've built support for OCSP and CRLs into my Rails app but unfortunately at least OCSP doesn't seem to work. I just get this error:
NameError: uninitialized constant OpenSSL::OCSP
Which leads me to believe it's not implemented - is that correct?
##

Issue #119, opened 2016-12-20
OpenSSL::ASN1 der translation inconsistent


HoneyryderChuck:
`#to_der` doesn't seem to be inconsistent encoding context-specific ASN1 data. This is a gist which hopefully demonstrate the issue:

```ruby
require "openssl"                                           
int = OpenSSL::ASN1::Integer.new(1)                                   
data = OpenSSL::ASN1::ASN1Data.new([int], 1, :CONTEXT_SPECIFIC)       
puts "original asn:"                                                  
puts data.inspect                                                     
der = data.to_der                                                    
puts "der is: #{der.unpack("H*")}"                           
asn = OpenSSL::ASN1.decode der
puts "asn:"                                                           
puts asn.inspect                                                      
```

Here is the output against jruby 9.1.6.0 (latest stable as of now):

```
original asn:
#<OpenSSL::ASN1::ASN1Data:0x7cc55834 @tag=1, @tag_class=:CONTEXT_SPECIFIC, @value=[#OpenSSL::ASN1::Integer:0x37a54f5e @tag=2, @tag_class=:UNIVERSAL, @tagging=nil, @value=1, infinite_length=false>]>
der is: ["810101"]
asn:
#<OpenSSL::ASN1::ASN1Data:0x434a314e @tag=1, @tag_class=:CONTEXT_SPECIFIC, @value=[#OpenSSL::ASN1::OctetString:0x4811fb9e @tag=4, @tag_class=:UNIVERSAL, @tagging=nil, value="\x01", @infinite_length=false>]>
```

which shows that the original asn can't be encoded and decoded into the same structure. 

Here's the output with MRI 2.1:

```
original asn:                                                                                                         
#<OpenSSL::ASN1::ASN1Data:0x007fb5b8e8dcb0 @tag=1, @value=[#OpenSSL::ASN1::Integer:0x007fb5b8e8dd00 @tag=2, @value=1, @tagging=nil, @tag_class=:UNIVERSAL, @infinite_length=false>], @tag_class=:CONTEXT_SPECIFIC, infinite_length=false> 
der is: ["a103020101"]                                                                                                
asn:                                                                                                                  
[...]

PR #124, opened 2017-03-08
add OCSP support

lampad:
An attempt to resolve https://github.com/jruby/jruby-openssl/issues/24

Finally found enough random time to finish this up.

Some assertions in test_ocsp.rb (ripped straight from 2.4.0 CRuby) are commented out because of some odd behavior I noticed with decoding objects using ASN1.decode. A cursory glance seems to suggest that bits of data are left out of the decoded objects or aren't put back in to the der string when calling to_der. 

I decided to TODO them since it felt like scope creep, and my mind just wasn't ready to dive into that side of things.

Let me know if you have questions or anything. This is my first real foray into JRuby, so if I did stupid stuff, please let me know.


headius:
Wow!

src/main/java/org/jruby/ext/openssl/OCSP.java
```diff
@@ -0,0 +1,211 @@
...
+    public static String getResponseStringForValue(IRubyObject fixnum) {
+        RubyFixnum rubyFixnum = (RubyFixnum) fixnum;
+        return responseMap.get((int)rubyFixnum.getLongValue());
+    }
+    
+    public static RaiseException newOCSPError(Ruby runtime, Exception ex) {
+        return Utils.newError(runtime, _OCSP(runtime).getClass("OCSPError"), ex);
+    }
+
+    public static RaiseException newSSLError(Ruby runtime, String message) {
```
kares: :name_badge: seems wrong

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
```diff
@@ -0,0 +1,147 @@
...
+* 
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+* FOR APARTICULAR PURPOSE AND NONINFRINGEMENT.
+* IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+* DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+* ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+* DEALINGS IN THE SOFTWARE.
+*
+*  Copyright (C) 2007-2009 Ola Bini <ola.bini@gmail.com>
```
kares: would prefer to have your name here instead of Ola's or just the `(C) 2017 The JRuby Team` if your shy :)

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
```diff
@@ -0,0 +1,147 @@
...
+    };
+    
+    public static void createBasicResponse(final Ruby runtime, final RubyModule _OCSP) {
+        RubyClass _BasicResponse = _OCSP.defineClassUnder("BasicResponse", runtime.getObject(), BASICRESPONSE_ALLOCATOR);
+        _BasicResponse.defineAnnotatedMethods(OCSPBasicResponse.class);
+    }
+    
+    private RubyString der;
+    private byte[] nonce;
+    private SecureRandom random = new SecureRandom();
```
kares: we should delay the `SecureRandom` creation - and use the helper from `SecurityHelper` (this is minor, really)

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
```diff
@@ -0,0 +1,147 @@
...
+            RubyString input = (RubyString)args[0];
+            nonce = generateNonce(input.getBytes());
+        }
+        
+        return this;
+    }
+    
+    @JRubyMethod(name = "add_status")
+    public OCSPBasicResponse add_status(final ThreadContext context) {
+        //TODO: Implement
```
kares: shouldn't we also print a warning on these stub impls? (am not sure - thus asking, there's helpers for that)

src/main/java/org/jruby/ext/openssl/OCSPRequest.java
```diff
@@ -0,0 +1,169 @@
...
+    public IRubyObject sign(IRubyObject cert, IRubyObject key, IRubyObject[] args) {
+        //TODO finish
+        X509Cert rubyCert = (X509Cert) cert;
+        
+        return this;
+    }
+    
+    @JRubyMethod(name = "to_der")
+    public IRubyObject to_der() {
+        //TODO implement
```
kares: it definitely makes sense to have warnings on these e.g. 
`OpenSSL.warn(context, "WARNING: unimplemented method called: OCSP::Request#to_der");`

src/main/java/org/jruby/ext/openssl/OCSPRequest.java
```diff
@@ -0,0 +1,169 @@
...
+        return RubyString.newEmptyString(getRuntime());
+    }
+    
+    @JRubyMethod(name = "verify")
+    public IRubyObject verify(IRubyObject certificates, IRubyObject store, IRubyObject flags) {
+        //TODO implement
+        
+        return RubyBoolean.newBoolean(getRuntime(), true);
```
kares: safer to return `false` if its unimplemented for users relying on this feature, right?

src/main/java/org/jruby/ext/openssl/StringHelper.java
```diff
@@ -29,6 +29,8 @@
 import java.util.ArrayList;
 import java.util.Locale;
 
+import org.bouncycastle.asn1.ASN1InputStream;
+import org.bouncycastle.asn1.ASN1Primitive;
```
kares: :no_entry:

kares:
CI also catched a (minor) issue :
```
[INFO] java.lang.ClassCastException: org.bouncycastle.jcajce.provider.asymmetric.x509.X509CertificateObject cannot be cast to org.bouncycastle.jce.provider.X509CertificateObject
[INFO] 	at org.jruby.ext.openssl.OCSPRequest.findCertByName(OCSPRequest.java:389)
[INFO] 	at org.jruby.ext.openssl.OCSPRequest.verify(OCSPRequest.java:307)
``` 

.. otherwise this is looking :grapes: as far as I can tell :cupid:

src/main/java/org/jruby/ext/openssl/OCSP.java
lampad: Yeah, not sure why I added this in the first place. Thanks for catching.

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
lampad: Haha, ok. I realized I need to add the license to the other files too.

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
lampad: The final version ended up not needing SecureRandom anyway

src/main/java/org/jruby/ext/openssl/OCSPBasicResponse.java
lampad: Sorry, those were notes for myself to implement them before pushing. The final version should have everything implemented. I could rebase/squash my commits if that would help so people wouldn't even see this. Thoughts?

src/main/java/org/jruby/ext/openssl/StringHelper.java
lampad: Woops!

lampad:
I'll update based on comments and try to figure out the CI failure.

lampad:
@kares I'm not sure what's going on with the build failures now.

One appears to have timed out: https://travis-ci.org/jruby/jruby-openssl/jobs/209400174

The others seem to be having trouble getting their environment/dependencies set up to run the ruby tests in their respective environments: https://travis-ci.org/jruby/jruby-openssl/jobs/209400177

I can't tell if I caused this problem or not... When I look at the file(s) in the stack in my checked out copy (jruby-openssl/lib/openssl.rb and jruby-openssl/lib/jopenssl/load.rb) i don't see the relevant lines... Am I missing something?

kares:
@lampad no worries. thanks for all the code!

HoneyryderChuck:
I've bundled the master branch, and I'm still seeing the same error:

```
NameError: uninitialized constant OpenSSL::OCSP
```

HoneyryderChuck:
@lampad , your issues might have to do with #119

lampad:
@HoneyryderChuck I'm not sure what "bundling" the master branch does exactly, but you'll probably need to package/repackage the new jopenssl.jar and the jruby-openssl gem and install the gem in your local gem env. There are instructions on how to do the packaging in the BUILDING.md file which should spit out a gem in the pkg dir.

lampad:
@HoneyryderChuck and yeah, I think you're correct re: #119. I didn't dig too deep into it there, but I did notice some functionality missing in the ASN1 class. If I find some time, I'll take a deep dive.

HoneyryderChuck:
@lampad I think bundler already takes care of that. `$LOAD_PATH` is at least pointing to a local bundled jruby-openssl directory (`.../jruby-openssl-580ea57f016b/lib`, which hash-key is the current last commit from master).

HoneyryderChuck:
here's the relevant bundler log run in verbose mode:

```
Fetching https://github.com/jruby/jruby-openssl.git
...
Using jruby-openssl 0.9.21.dev (java) from https://github.com/jruby/jruby-openssl.git (at master@580ea57)
0:  jruby-openssl (0.9.21.dev) from /Users/tiagocardoso/Projects/jaguar/.bundle/jruby/2.3.0/bundler/gems/jruby-openssl-580ea57f016b/jruby-openssl.gemspec
...
```

kares:
@HoneyryderChuck jruby-openssl does not work correctly when using `:git => ...` native-ext compilation won't happen.

HoneyryderChuck:
@kares thx for the info. That's unfortunate, my expectation for bundler would be definitely something else. Is this a pending/in-progress feature, and does it concern bundler or jruby?

After downloading the package, is there a way to automate installing the local package with bundler, or do I resort to following the instructions in BUILDING.md, as suggested by @lampad ?

kares:
that's a usual issue with native gems as they require building (e.g. same with nokogiri), this could be hacked to get working in an 'un-official' way. it does not concern JRuby nor Bundler atm.

you just gem install the downloaded gem and than use it (specify its version in *Gemfile*). this is a bit out-of-scope here. just do the same thing as you would if you had a third-party gem not on RubyGems.

HoneyryderChuck:
@kares, doh, you're right :) managed to do it, and on to the next issue.

Issue #125, opened 2017-03-13
Few cipher suites on IBM JRE


ysohda:
Hi all,

There are a few cipher suites on IBM JRE even though IBM JRE supports more ciphers.
Here are results of cipher suites on JRuby 9.1.8.0 with IBM JRE 8.0.4.1 (and Oracle JRE 1.8.0_121).

JRuby 9.1.8.0 IBM JRE 8.0.4.1 (/w unrestricted JCE policy)
>jruby  -e "require 'openssl'; ctx=OpenSSL::SSL::SSLContext.new; ctx.ciphers.each { |c| p c}"
["EXP-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["EXP-EDH-RSA-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["EXP-EDH-DSS-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["DES-CBC-SHA", "TLSv1/SSLv3", 56, 56]
["EDH-RSA-DES-CBC-SHA", "TLSv1/SSLv3", 56, 56]

Oracle JDK 1.8.0_121 (/w unrestricted JCE policy)
>jruby  -e "require 'openssl'; ctx=OpenSSL::SSL::SSLContext.new; ctx.ciphers.each { |c| p c}"
["EXP-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["EXP-EDH-RSA-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["EXP-EDH-DSS-DES-CBC-SHA", "TLSv1/SSLv3", 40, 56]
["DES-CBC-SHA", "TLSv1/SSLv3", 56, 56]
["EDH-RSA-DES-CBC-SHA", "TLSv1/SSLv3", 56, 56]
["ECDHE-ECDSA-AES128-SHA256", "TLSv1/SSLv3", 128, 256]
["ECDHE-RSA-AES128-SHA256", "TLSv1/SSLv3", 128, 256]
["ECDH-ECDSA-AES128-SHA256", "TLSv1/SSLv3", 128, 256]
["ECDH-RSA-AES128-SHA256", "TLSv1/SSLv3", 128, 256]
["ECDHE-ECDSA-AES128-SHA", "TLSv1/SSLv3", 128, 128]
["ECDHE-RSA-AES128-SHA", "TLSv1/SSLv3", 128, 128]
[...]

PR #126, opened 2017-03-13
Add cipher suite strings for IBM JRE

ysohda:
Add cipher suite strings for IBM JRE into SuiteToOSSL mapping

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/CipherStrings.java
+++ b/src/main/java/org/jruby/ext/openssl/CipherStrings.java
@@ -1755,6 +1755,12 @@ public class CipherStrings {
             SSL_NOT_EXP|SSL_HIGH, 128, 256, SSL_ALL_CIPHERS, SSL_ALL_STRENGTHS
         ));
 
+        SuiteToOSSL.put("TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384", name = "ECDHE-ECDSA-AES256-SHA384");
+	    CipherNames.put(name, new Def(name,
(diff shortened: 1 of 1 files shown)


kares:
Thanks Yuki, this is excellent ... if you have a change could you update the comment with a note from #125 

from :
```
/* For IBM JRE */
```
to something like :
```
/* For IBM JRE: suite names start with "SSL_". On Oracle JRE, the suite names start with "TLS_" */
```

ysohda:
Hi Kares,

Thank you for your review.
I updated the comment and also made EDCH/EDCHE SHA384 cipher suites available not only for IBM JRE but also for Oracle JRE.
Btw, about AES GCM ciper cuites introduced in Java 8, I added mapping entries for IBM JRE. So if AES GCM cipher suites are available on jruby-openssl, they will be also available on IBM JRE.

Thank you,
Yuki.

ysohda:
Hi @kares ,

Can I have any update?
It would be nice that the fix is included in the next JRuby-Openssl and JRuby releases (0.9.21 & 9.1.9).

Thank you,
Yuki.

kares:
Sure will be in next jossl release. Not sure if it will make it into 9.1.9 (its a lot of boilerplate to test a release).
But you'll be able to `gem update` later and we can push you a dev SNAPSHOT if that is usable for you ...

kares:
https://oss.sonatype.org/content/repositories/snapshots/rubygems/jruby-openssl/0.9.21.dev-SNAPSHOT/

ysohda:
Hi Kares,

Thank you for creating a snapshot. I'm going to use it.
May I have an approximate next release date if it is planned?

Thank you,
Yuki.

PR #136, opened 2017-06-11
move "DEFAULT" handling right into CipherStrings

MSNexploder:
Move "DEFAULT" special case handling further down and match OpenSSL behaviour
This fixes https://github.com/jruby/jruby/issues/2193


kares:
looking good, thanks!

Issue #134, opened 2017-05-22
X509ExtensionFactory generates incorrect extension for subjectAltName


roadrunner2:
The extension generated for `subjectAltName` by `X509ExtensionFactory` is missing a sequence. Take this code snippet:
```
extensions = OpenSSL::X509::ExtensionFactory.new
ext = extensions.create_extension("subjectAltName", "email:foo@bar.com,DNS:a.b.com")
File.open("/tmp/san.ext", "w") { |f| f.print(ext.to_der) }
```

The DER of this extension should look like (and does so under MRI)
```
   0 30   31: SEQUENCE {
   2 06    3:   OBJECT IDENTIFIER subjectAltName (2 5 29 17)
   7 04   24:   OCTET STRING, encapsulates {
   9 30   22:       SEQUENCE {
  11 81   11:         [1] 'foo@bar.com'
  24 82    7:         [2] 'a.b.com'
            :         }
            :       }
            :   }
```

But the actual DER of the created extension under JRuby is
```
   0 30   32: SEQUENCE {
   2 06    3:   OBJECT IDENTIFIER subjectAltName (2 5 29 17)
   7 04   25:   OCTET STRING, encapsulates {
   9 81   23:       [1] 'foo@bar.com,DNS:a.b.com'
            :       }
            :   }
```
Note the missing sequence, and the fact that both values are in one string. 

The core issues are that ```X509ExtensionFactory.parseSubjectAltName()``` returns a GeneralName instead of a GeneralNames (sequence of GeneralName), and that it fails to parse multiple names properly.

Due to the missing sequence, it's currently completely impossible to generate a (valid) certificate with a subject-alt-name extension.

Lastly, pull request #123 appears to be related to this.

kares:
please give 0.9.21 a try a let us know if there's smt wrong. assuming its fixed here.

duritong:
Unfortunately, the problem with #123 still persists, right? So the representation of SubjectAltNames is still flawed, though this problem might be hidden now...

kares:
@duritong its a different issue (and that is why its still open) that would take more time which I did not have.
am grateful for any help as progress is otherwise low with only 1 person managing the project forward occasionally (also have other priorities atm when eventually I get work done on jossl)

roadrunner2:
@kares Thanks for looking into this. While your change appears to fix one very special case, the parsing is still fairly broken if the string doesn't happen to list email first, or it contains just a URI, etc. E.g. all of these still fail:
[...]

PR #140, opened 2017-08-05
Subject alt name parsing fixes

roadrunner2:
Pull with fixes for subject-alt-name parsing, as requested in #134.


kares:
really not sure about this piece - seems like there's more indirection (incorrect parsing work-arounds) ;(
simply lack the time of looking at this properly, but my hopes are at least the new test piece ~ same on MRI

PR #143, opened 2017-09-26
Improve jvm version detection for java 9

ketan:


kares:
thanks its looking good. have planned on doing some cleanup (and drop some old Java/JRuby support) for next version 0.10 ... this would be a perfect fit to include. however I no concrete plans (yet) for this as I am a bit overloaded. hopefully you won't mind if Java 9 support isn't released right away...

ketan:
I saw some commits about dropping java 1.6 support, but could see some 1.6 specific code (jvm detection at the very least) that can probably be removed. If you could merge this PR, I'm happy to submit an additional PR to get rid of some more cruft.

ketan:
I don't mind you holding this PR for a few more days, but it's a blocker for us because it prevents us from using JRuby 9000 on java 9 by not being able to do even a basic `gem install`.

When do you think you can release (no pressure :))

src/main/java/org/jruby/ext/openssl/OpenSSL.java
```diff
@@ -255,27 +255,27 @@ static void warn(final ThreadContext context, final IRubyObject msg) {
     public static String javaVersion(final String def) {
         final String javaVersionProperty =
                 SafePropertyAccessor.getProperty("java.version", def);
-        if ( javaVersionProperty == "0" ) return "1.7.0"; // Android
+        if (javaVersionProperty.equals("0")) return "1.7.0"; // Android
```
kares: this change makes it no longer null-safe as it was before - could you please swap the equals comparison, just in case

src/main/java/org/jruby/ext/openssl/OpenSSL.java
```diff
@@ -255,27 +255,27 @@ static void warn(final ThreadContext context, final IRubyObject msg) {
...
-        final int gt = "1.8".compareTo( javaVersion("0.0").substring(0, 3) );
+        final int gt = new Version("1.8").compareTo(new Version(javaVersion("0.0")));
         return atLeast ? gt <= 0 : gt == 0;
     }
 
     static boolean javaVersion9(final boolean atLeast) {
-        final int gt = "9".compareTo( javaVersion("0").substring(0, 1) );
+        final int gt = new Version("9").compareTo(new Version(javaVersion("0.0")));
         return atLeast ? gt <= 0 : gt == 0;
     }
```
kares: maybe also inner `static class Version implements ...` here, 
prefer not to have 'public' utils (internally) used in one place

ketan:
Can we get a release out, so I can submit a PR to jruby to bump jruby-openssl?

kares:
we're planning on managing that out but we would like to find out why we have [these Java 9 mvn failures](https://travis-ci.org/jruby/jruby-openssl/jobs/294739803)

mkristian:
@kares I pushed mavengem wagon to https://oss.sonatype.org/content/repositories/snapshots/

use it by adding to Mavenfile:
```
  extension 'org.torquebox.mojo:mavengem-wagon:1.0.0-SNAPSHOT'
```

this is build using jruby-9.1.14.0 with the SNAPSHOT of jruby-openssl you deployed yesterday.

I did see this `java.lang.ExceptionInInitializerError` this morning this 1.0.0-SNAPSHOT fixed it for me.

kares:
@mkristian perfect - thanks, let's give it a try and than I think we'll push a release into staging 
... to see how its doing in JRuby's repo running MRI OpenSSL test

ketan:
❤️

perlun:
I started seeing this issue now with Java 10:

```
$ bundle exec rake -T
rake aborted!
LoadError: load error: jopenssl/load -- java.lang.StringIndexOutOfBoundsException: begin 0, end 3, length 2
/Volumes/git/ecraft/ecraft.uxfactory.server/Rakefile:1:in `block in (root)'
```

Has there been a release of `jruby-openssl` with this feature included? I have been running JRuby on Java 9 without (big) problems, so I wonder why this has now reappeared on Java 10.

perlun:
Debugged this a bit more, filed a new issue - #157.

PR #146, opened 2017-10-11
Normalize all constants in CipherStrings as public final static

drcapulet:


kares:
thx, do you need to depend on these? (we could use PR's description on the reasons behind the change)
generally like to expose as little as possible, but if its useful (seems so) than o.k.

drcapulet:
No technical reason for these being public, just normalized them to make the code a little more readable.

PR #149, opened 2017-11-02
Replace incoming certs with newer versions from trust store.

headius:
Fixes jruby/jruby#4802

This tries to emulate what OpenSSL does in X509_vfy.c when building the untrusted certificate chain, replacing incoming certs with newer versions from the local store.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/x509store/StoreContext.java
+++ b/src/main/java/org/jruby/ext/openssl/x509store/StoreContext.java
@@ -28,6 +28,7 @@
 package org.jruby.ext.openssl.x509store;
 
 import java.security.GeneralSecurityException;
+import java.security.Principal;
 import java.security.PublicKey;
 import java.security.cert.X509CRL;
 import java.security.cert.X509Certificate;
@@ -38,12 +39,15 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.ListIterator;
 import java.util.Set;
 
 import org.bouncycastle.asn1.ASN1InputStream;
 import org.bouncycastle.asn1.ASN1Integer;
 import org.bouncycastle.asn1.ASN1Sequence;
 import org.jruby.ext.openssl.SecurityHelper;
+
+import javax.security.auth.x500.X500Principal;
 
 /**
  * c: X509_STORE_CTX
@@ -661,6 +665,22 @@ public class StoreContext {
         List<X509AuxCertificate> sktmp = null;
         if ( untrusted != null ) {
             sktmp = new ArrayList<X509AuxCertificate>(untrusted);
+
+            // replace certs in untrusted with trusted versions if found
+            X509Object[] objTmp = {null};
+            for (ListIterator<X509AuxCertificate> iter = sktmp.listIterator(); iter.hasNext();) {
+                X509AuxCertificate skCert = iter.next();
+                X500Principal principal = skCert.cert.getSubjectX500Principal();
+                int ok = getBySubject(X509Utils.X509_LU_X509, new Name(principal), objTmp);
+                if (ok == X509Utils.X509_LU_X509) {
+                    // replace old with new and clear rest of untrusted
+                    iter.set(((Certificate) objTmp[0]).x509);
+                    while (iter.hasNext()) {
+                        iter.next();
+                        iter.remove();
+                    }
+                }
+            }
         }
         num = chain.size();
         x = chain.get(num - 1);


headius:
The logic here is derived from pieces of logic in [OpenSSL's x509_vfy.c build_chain function](https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vfy.c#L2830), specifically [this logic removing remaining untrusted certs](https://github.com/openssl/openssl/blob/master/crypto/x509/x509_vfy.c#L2998) and [this logic replacing a wire cert with one from the trust store].

headius:
Tagging a couple other folks that might be able to help: @nahi @tarcieri

headius:
Also relevant, the discussion about why this change was made in OpenSSL: http://openssl.6102.n7.nabble.com/openssl-dev-openssl-org-2634-Cross-signed-certs-rejected-by-OpenSSL-because-root-cert-not-base-of-chn-td56649.html

src/main/java/org/jruby/ext/openssl/x509store/StoreContext.java
```diff
@@ -661,6 +665,26 @@ public int verifyCertificate() throws Exception {
...
+            // replace certs in untrusted with trusted versions if found
+            X509Object[] objTmp = {null};
+            for (ListIterator<X509AuxCertificate> iter = sktmp.listIterator(); iter.hasNext();) {
+                X509AuxCertificate skCert = iter.next();
+                X500Principal principal = skCert.cert.getSubjectX500Principal();
+                int ok = getBySubject(X509Utils.X509_LU_X509, new Name(principal), objTmp);
+                if (ok == X509Utils.X509_LU_X509) {
+                    // replace old with new and clear rest of untrusted
+                    Certificate certificate = (Certificate) objTmp[0];
+                    if (certificate.x509.equals(skCert)) {
```
mkristian: I am sure I am missing bits and pieces. the question for me is what cert-chain do I need to send to get it trusted. so I need the `skCert` to match the subject-principal of a trusted certificate and then I need to satisfy this equals method.

from the supplied threads I figured that we basically want to replace self-signed certificates in the middle of the chain by some trusted certificate from local trust-store.

headius: @mkristian Yes, that's exactly it. I'm working on getting the right things to compare to allow an untrusted cert from the wire to properly match a trusted cert in the local store. Need to read more of the OpenSSL code to understand exactly how much needs to be the same.

mkristian: @headius I was pounding a bit more on this. maybe this is just OK as is when there new trusted certificate chain gets verified after this. after all that is what we want to achieve: a trusted verified chain from the server certificate to root certificate where root comes from our root certificate store.

original-brownbear:
@headius can I do anything to help move this along?
The lack of stable Java9 support is becoming a big issue for Logstash and I'm happy to help in any way I can to move this forward :)
The patch here makes sense to me in an of itself.
But for Java9 in general, it looks like there's also a build issue with the Maven GEM  wagon (that depends on JRuby) not working on Java9 before this is merged + released, making a proper Java9 test build impossible?

=> If you have any concrete steps in mind for resolving this situation, I'm happy to offer my time for that :)

kares:
@original-brownbear this was an attempt to resolve a (system) CA cert issue.
at the time of the patch it didn't simply reproduce reliably (maybe related to ca.cert update on nix systems), but now there are test failures and those occur with or without this patch. 
we can work-around the wagon build issue eventually, so I wouldn't worry about that much as release is blocked by that cert verification failure - not sure if it makes sense to release jossl as is (likely not).

original-brownbear:
@kares I was able to reproduce the failure with Oracle `build 9.0.4+11` but was able to reproduce it with `9.0.1`. It seems there have been a bunch of fixes the the cert management in `9.0.4+` that seem related http://www.oracle.com/technetwork/java/javase/9-0-4-relnotes-4021191.html.

Maybe we should try upping Travis to `9.0.4` and see how that goes build wise?

kares:
@original-brownbear thanks, still Java 8 is the primary target. recall trying latest at some point wout luck.
not sure if smt changed since - maybe worth a re-check. 
if its resolved (test :green_heart:) we need to find out which version and issue a warning for users running 'old' 8.

headius:
Anyone checked with a more recent Java 8? If it got fixed in 9.0.4 it may have been backported too.

albertogg:
Hi, this is still happening in alpine as of version:

```
jruby 9.1.15.0 (2.3.3) 2017-12-07 929fde8 OpenJDK 64-Bit Server VM 25.151-b12 on 1.8.0_151-b12 +jit [linux-x86_64]
===```

kares:
yeah, I recalled and revisited this.
but the change presented here won't cut it, at least not for the "Let's Encrypt" case (already verified that).
anyhow, I was on the path of porting the whole OpenSSL 1.1 logic but than got into revisiting the 1.0 logic (under the assumption that having `X509_V_FLAG_TRUSTED_FIRST` would be sufficient - it isn't). 
so I am back looking at `build_chain`, which also motivated this PR, while also trying to minimize the changeset to be able to push a fix in a patch release.

PR #155, opened 2018-02-16
Don't use padding for streaming cipher modes

### LINKED ISSUE #13: Padding differences against MRI for cipher modes such as OFB
I'm working on an RPC interface at the moment with the POC piece in Ruby, but I wrote a Java client.

I found that using bf-ofb cipher that MRI Ruby used no padding, not going to get into it should or it shouldn't but the current default for MRI is not to pad with the bf-ofb Cipher, hence my JCE implementation uses Cipher.getInstance("Blowfish/OFB64/NoPadding") and this works quite happily with MRI Ruby on default settings and no cipher.padding = 0 added.

The documentation is a little vague but the behaviour is definitely there that padding isn't added by MRI but on JRuby it seems to default the other way.

I did find an artical, way back (can't find it now) that states that OFB shouldn't use padding but I'm not going to get into that since this seems to be the behaviour for MRI Ruby.

I've just been more explicit in my code and added cipher.padding = 0 to my code but if you are interested, for these modes, theres a subtle different.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6383790-padding-differences-against-mri-for-cipher-modes-such-as-ofb?utm_campaign=plugin&utm_content=tracker%2F136995&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F136995&utm_medium=issues&utm_source=github).
</bountysource-plugin>

dgolombek:
OFB, CFB[8], CTR, and GCM cipher modes don't require padding, since they
act in a streaming manner, working byte-by-byte. Adding padding to them
makes the output incompatible with MRI, and unable to be decrypted with
it (and OpenSSL, underneath it).

GCM is added to NO_PADDING_BLOCK_MODES despite not being in
KNOWN_BLOCK_MODES to keep backward compatibility in getPaddingType. I'm
happy removing it if others agree, since there shouldn't be any way for
it to be supported currently.

Fixes #13

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/Cipher.java
+++ b/src/main/java/org/jruby/ext/openssl/Cipher.java
@@ -261,6 +261,10 @@ public class Cipher extends RubyObject {
             KNOWN_BLOCK_MODES.add("NONE"); // valid to pass into JCE
         }
 
+        // Subset of KNOWN_BLOCK_MODES that do not require padding (and shouldn't have it by default).
+        private static final List<String> NO_PADDING_BLOCK_MODES = Arrays.asList(
+            "CFB", "CFB8", "OFB", "CTR", "GCM");
+
         // Ruby to Java name String (or FALSE)
         static final HashMap<String, String[]> supportedCiphers = new LinkedHashMap<String, String[]>(120, 1);
         // we're _marking_ unsupported keys with a Boolean.FALSE mapping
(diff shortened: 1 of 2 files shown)


kares:
Thanks Dave.

Issue #161, opened 2018-05-03
Warning: Problem with Java 9, 10


lberoiza:
There are warnings by using Java9 or Java10

___________________________________________________________

**WARNING: Illegal reflective access by org.jruby.ext.openssl.SecurityHelper (file:/home/amos/Amos/Project_Russia/common/smile/rvm/rubies/jruby-9.1.17.0/lib/ruby/stdlib/jopenssl.jar) to constructor java.security.cert.CertificateFactory(java.security.cert.CertificateFactorySpi,java.security.Provider,java.lang.String)**
	at org.jruby.ext.openssl.SecurityHelper.newInstance(SecurityHelper.java:794)
	at org.jruby.ext.openssl.SecurityHelper.getCertificateFactory(SecurityHelper.java:249)
	at org.jruby.ext.openssl.SecurityHelper.<clinit>(SecurityHelper.java:152)
	at org.jruby.ext.openssl.OpenSSL.createOpenSSL(OpenSSL.java:57)
	at org.jruby.ext.openssl.OpenSSL.load(OpenSSL.java:49)
	at org.jruby.javasupport.JavaMethod.invokeDirectWithExceptionHandling(JavaMethod.java:453)
	at org.jruby.javasupport.JavaMethod.invokeStaticDirect(JavaMethod.java:365)
	at org.jruby.java.invokers.StaticMethodInvoker.call(StaticMethodInvoker.java:47)
	at org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:318)
	at org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:155)
	at org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:315)
	at org.jruby.ir.interpreter.StartupInterpreterEngine.interpret(StartupInterpreterEngine.java:72)
	at org.jruby.ir.interpreter.Interpreter.INTERPRET_ROOT(Interpreter.java:107)
	at org.jruby.ir.interpreter.Interpreter.execute(Interpreter.java:94)
	at org.jruby.ir.interpreter.Interpreter.execute(Interpreter.java:34)
	at org.jruby.ir.IRTranslator.execute(IRTranslator.java:42)
	at org.jruby.Ruby.runInterpreter(Ruby.java:842)
	at org.jruby.Ruby.loadFile(Ruby.java:2903)
[...]

PR #162, opened 2018-05-03
[issues/161] Comment function not allowed Java (9,10) > 8

lberoiza:
Possible Solution to Issue #161


kares:
sorry, but this isn't a solution really - Java only prints you with the first warning.
after this you might likely be getting more + as CI suggest not setting the ctor accessible breaks stuff!

lberoiza:
Hi,

Do you have a suggestion, how we can solve this issue, I don't really understand what you mean. 
Regards
Luis

mkristian:
@lberoiza not setting `constructor.setAccessible(true);` breaks functionality (see travis-ci ). so unless we find a way to retain functionality and remove the warning we probably stick with the warning.

kares:
hmm so this is going in a way of silencing the JDK internal logger **reflectively** :)
very much appreciate all the effort ... but not sure we would want code such as these around.
it doesn't really solve the issue + we would probably need a fork treating unsafe as JRuby does etc.
long term solution would be to rewrite most of these to use light-weight BC APIs and do not rely on JDK APIs

kares:
we've attempted to reduce the reflection performed by security helper (for 0.10.5), if this turns out problematic we should attempt some silencing but generally we do not want to have "ugly" module workarounds. thanks for your code thought, very interesting that people actually use these hacks.

PR #166, opened 2018-05-31
Fix jopenssl24 set_key methods

jboler:
With jRuby 9.2.0.0 which targets Ruby 2.5 I'm getting:

    NoMethodError: undefined method `public_key=' for #<OpenSSL::PKey::DSA:0x64b83f76>

It works fine when I use `pub_key=`


jboler:
This bug will now manifest on jRuby 9.1.18.0 if it's released without this fix. Previously the bug wasn't loaded on 9.1 because of `load 'jopenssl24.rb' if RUBY_VERSION >= '2.4'`.

PR #170, opened 2018-07-21
additional buffer argument

ahorek:
https://github.com/jruby/jruby/issues/5242

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/Cipher.java
+++ b/src/main/java/org/jruby/ext/openssl/Cipher.java
@@ -68,6 +68,7 @@ import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.jruby.runtime.Visibility;
 import org.jruby.util.ByteList;
+import org.jruby.util.TypeConverter;
 
 import static org.jruby.ext.openssl.OpenSSL.*;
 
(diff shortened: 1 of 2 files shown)


src/test/ruby/test_cipher.rb
```diff
@@ -475,4 +475,19 @@ def test_encrypt_aes_cfb_20_incompatibility
...
   end
 
+  def test_encrypt_aes_256_cbc_modifies_buffer
+    cipher = OpenSSL::Cipher.new("AES-256-CBC")
+    cipher.key = "a" * 32
+    cipher.encrypt
+    buffer = ''
+    actual = cipher.update('bar' * 10, buffer)
+    if jruby?
+      expected = "\xE6\xD3Y\fc\xEE\xBA\xB2*\x0Fr\xD1\xC2b\x03\xD0"
```
ahorek: maybe a bug?

src/main/java/org/jruby/ext/openssl/Cipher.java
```diff
@@ -1104,6 +1104,11 @@ private String getCipherAlgorithm() {
 
     @JRubyMethod
     public IRubyObject update(final ThreadContext context, final IRubyObject arg) {
+        return update(context, arg, null);
+    }
+
+    @JRubyMethod
+    public IRubyObject update(final ThreadContext context, final IRubyObject arg, IRubyObject buffer) {
```
enebo: As a @JRubyMethod buffer can be any Ruby type.  So you need to do something like TypeConverter.checkStringType.

ahorek: thanks

src/test/ruby/test_cipher.rb
enebo: @ahorek I just saw this.  I don't know.  I would expect the two impls to generate the same data.

janko-m:
Thanks for the fix! In which JRuby version is this going to be released?

PR #177, opened 2018-11-15
Add ECDHE-RSA-AES128-GCM-SHA256 support

mattbooks:
This adds support for ECDHE-RSA-AES128-GCM-SHA256. Many more are still missing,
but I am not familiar enough with openssl to go through and configure them.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/CipherStrings.java
+++ b/src/main/java/org/jruby/ext/openssl/CipherStrings.java
@@ -1794,6 +1794,12 @@ public class CipherStrings {
             SSL_NOT_EXP|SSL_HIGH, 128, 256, SSL_ALL_CIPHERS, SSL_ALL_STRENGTHS
         ));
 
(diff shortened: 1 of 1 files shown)


betesh:
@mattbooks I might have to fork this.  Do you mind rebasing onto master?

@kares Anything holding up merging this?

kares:
a binary *jopenssl.tar.gz* file ...
+ was hoping someone could review the full suite instead of adding ciphers 1-by-1

betesh:
> a binary jopenssl.tar.gz file ...

I see one in the PR.  Not sure I understand?

mattbooks:
It's been a while, but the tarball is probably a mistake and does not need to be included.

Updating the full suite is definitely a good idea but not something I'm able to do. This was a workaround to get me out of a bind, which I no longer am using.

benissimo:
I've removed the tarball and rebased onto latest master here:
https://github.com/jruby/jruby-openssl/pull/185

ThomasKoppensteiner:
What is the status of this PR?
Should it be closed and we continue with PR #185 ?

PR #198, opened 2020-04-06
Allow multiple Certificates with the same SubjectDN in the store

jpsikorra:
## Description
The current implementation of [OpenSSL::X509::Store](https://ruby-doc.org/stdlib-2.4.0/libdoc/openssl/rdoc/OpenSSL/X509/Store.html) in jruby-openssl does not allow multiple X509Certificates with the same SubjectDN (see [Certificate.java#L57](https://github.com/jruby/jruby-openssl/blob/6f91360cafa489ca23055bcdf5d49e6c56aa2018/src/main/java/org/jruby/ext/openssl/x509store/Certificate.java#L57)). This differs from the [C implementation](https://github.com/openbsd/src/blob/8d6bc8b3e8c3426e6a399728334939c60ac42a0d/lib/libcrypto/x509/x509_lu.c#L652). This may be tested by adding two certificates with the same SubjectDN to a OpenSSL::X509::Store and executing `verify` on certificate signed by one of them. It will work in CRuby and will not work in JRuby depending on the order the certificates are added.

## Changes
The `matches` method of Certificate was updated to compare on [hashCode()](https://docs.oracle.com/javase/7/docs/api/java/security/cert/Certificate.html#hashCode()). 
Because [getFirstIssuer()](https://github.com/jruby/jruby-openssl/blob/6f91360cafa489ca23055bcdf5d49e6c56aa2018/src/main/java/org/jruby/ext/openssl/x509store/StoreContext.java#L169) expects a grouping based on SubjectX500Principal of the certificate objects in the store, this was implemented, too.

## How this was tested
I tested if this change will allow the usage of multiple certificates with the same subjectdn on windows and linux. More tests are probably necessary.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/x509store/Store.java
+++ b/src/main/java/org/jruby/ext/openssl/x509store/Store.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.List;
 
 import javax.net.ssl.X509TrustManager;
+import javax.security.auth.x500.X500Principal;
 
 import org.jruby.Ruby;
 import org.jruby.ext.openssl.OpenSSL;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jruby/ext/openssl/x509store/Store.java
```diff
@@ -329,8 +330,25 @@ private synchronized int addObject(final X509Object xObject, final int prevLengt
...
+        int idx = length;
+        if (xObject instanceof Certificate) {
+            final X500Principal p1 = ((Certificate) xObject).x509.getIssuerX500Principal();
+            final Name n1 = new Name(p1);
+
+            for (idx = 0; idx < objects.length; idx++) {
+                X509Object xMember  = objects[idx];
+                if (xMember instanceof Certificate) {
+                    X500Principal p2 = ((Certificate) xMember).x509.getIssuerX500Principal();
+                    if(n1.equalTo(p2)) break;
```
kares, line 344: seems we're comparing Name to a X500Principal object, that does not seem right.

jpsikorra:
The insert after the same-named cert is done because of the way [getFirstIssuer](https://github.com/jruby/jruby-openssl/blob/6f91360cafa489ca23055bcdf5d49e6c56aa2018/src/main/java/org/jruby/ext/openssl/x509store/StoreContext.java#L169) works. It finds the first certificate with a matching subject. Then it starts from that index to look for a real match. As soon as the subject does not match, it returns. That implies that certificates in the store must be ordered by subject if multiple certificates with the same subject exist.

PR #203, opened 2020-06-25
Get all failing CI jobs report a failing CI status

### LINKED ISSUE #199: CI shows as green against java 11, but it's actually red
I found this during #195, but I figure I'll report it separately since it's a different issue.

See for example https://travis-ci.org/github/jruby/jruby-openssl/jobs/667216105.

The problem lies in the `rmvn` executable failing to report the right process status. The is simple, but it should be made inside either `ruby-maven or `ruby-maven-libs` code I believe.

I'll work on a fix for this.

deivid-rodriguez:
This PR fixes #199.

It uses the fix from https://github.com/jruby/ruby-maven/pull/1.

It also fixes some of the errors uncovered by that change.

Now a CI job should report a green status if and only if it's green.

This makes my problems with java 11 reported as #195 easily visible in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ matrix:
     - jdk: openjdk7
       env: TEST_PROFILE=test-9.1.17.0
     - jdk: oraclejdk11
-      env: TEST_COMMAND="jruby -rbundler/setup -S rake integration:install integration:test"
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This showed that currently 50% of CI jobs are green.

kares:
:green_heart:  David, want this so badly that I will merge depending on your fork :wink: please keep it up for a while.

deivid-rodriguez:
Hahaha, sure! Thanks for merging :)

PR #215, opened 2020-10-26
Fix OpenSSL::SSL::SSLContext#min_version= failure

MariuszCwikla:
This is attempt to fix jruby/jruby#6409.

I couldn't test it with unit tests due to other issues. However I tested it with following script 
[issue_6409.zip](https://github.com/jruby/jruby-openssl/files/5438096/issue_6409.zip)
I compared results between ruby and jruby and it seems it's working almost the same, e.g. 
* different error messages
* ruby fails when setting min_version to SSLv2
I believe these discrepancies could be ignored.

I attempted to implement unit test `test_minmax_ssl_version` however `server` method fails even on master with `no cipher match` error. So firstly master branch should be stabilized due to this and other issues before this PR can be finished.

This is still WIP, however comments are welcome.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/SSLContext.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLContext.java
@@ -53,6 +53,7 @@ import javax.net.ssl.X509TrustManager;
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
 import org.jruby.RubyClass;
+import org.jruby.RubyFixnum;
 import org.jruby.RubyHash;
 import org.jruby.RubyInteger;
 import org.jruby.RubyModule;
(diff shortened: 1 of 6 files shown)


djberg96:
Thanks for tackling this!

> ruby fails when setting min_version to SSLv2

Perhaps C Ruby is using whatever is set in the openssl config?

src/main/java/org/jruby/ext/openssl/SSLContext.java
```diff
@@ -517,17 +594,47 @@ public IRubyObject set_ssl_version(IRubyObject version) {
...
+        maxProtocolVersion = parseProtoVersion(maxVersion);
+
+        return context.nil;
+    }
+
+    private int parseProtoVersion(IRubyObject version) {
+        if (version.isNil())
+            return 0;
+        if (version instanceof RubyFixnum) {
+            return (int) ((RubyFixnum) version).getLongValue();
```
kares, line 624: let's just to `getIntValue()` here so we handle overflown values

src/test/ruby/ssl/test_socket.rb
```diff
@@ -136,6 +136,21 @@ def test_connect_nonblock
...
   end if RUBY_VERSION > '2.2'
 
+  def test_minmax_ssl_version
+    # WIP: this test should setup it's own server to prove that SSLContext.min_version and SSLContext.max_version work as expected
+    # However below line is failing  with "no cipher match" even on master
+    # When "no cipher match" is fixed, then this test need cleanup
+
+    #ssl_server = server         
+    begin
+      ssl_client = client(443, host: "google.com", min_version: OpenSSL::SSL::TLS1_VERSION, max_version: OpenSSL::SSL::TLS1_1_VERSION, ciphers: nil)
```
kares, line 146: believe we want some unit tests (done above except for the `set_minmax_proto_version`) for the version parsing/mapping instead of an integration tests here. 
also not sure hard-coding what google has on their servers atm. 
anyhow it's a good start but we might want to do an integration tests for this [elsewhere](https://github.com/jruby/jruby-openssl/tree/master/src/test/integration).

Gemfile
```diff
@@ -11,7 +11,7 @@ gem 'jar-dependencies', '~> 0.3.11', :require => nil
 gem 'mocha', '~> 1.4', '< 2.0'
 
 # for the rake task
-gem 'ruby-maven', github: 'deivid-rodriguez/ruby-maven', branch: 'exit_status'
+gem 'ruby-maven', github: 'deivid-rodriguez/ruby-maven', branch: 'master'
```
kares, line 14: please rebase, this was due https://github.com/jruby/ruby-maven/pull/1 which is merged upstream now

lib/jopenssl23/openssl/ssl.rb
```diff
@@ -214,7 +214,7 @@ def ssl_version=(meth)
           raise ArgumentError, "unknown SSL method `%s'" % meth
         set_minmax_proto_version(version, version)
         @min_proto_version = @max_proto_version = version
-      end unless method_defined? :ssl_version=
```
kares, line 217: let's keep this as is as it seems unrelated to the PR changes?

kares (commented):
:+1: good work, we need a couple of minor changes as outlined in the comments

src/main/java/org/jruby/ext/openssl/SSLContext.java
MariuszCwikla: Unfortunately there is no `getIntValue()`.
But I double checked and found `static int fix2int(RubyFixnum arg)`

lib/jopenssl23/openssl/ssl.rb
MariuszCwikla: I believe I had a reason to remove this. I compared this with CRuby version: https://github.com/ruby/ruby/blob/ruby_2_6/ext/openssl/lib/openssl/ssl.rb#L213
which does not have `unless method_defined? :ssl_version=`

I'll double check it and revert if it's fine.

src/test/ruby/ssl/test_socket.rb
MariuszCwikla: Yes, totally agree.

This PR will still need to be polished, more tests added.
But first I believe that master branch should be stabilized.

Gemfile
MariuszCwikla: Done

lib/jopenssl23/openssl/ssl.rb
MariuszCwikla: Ok, removed.
Only one smalll difference, unrealistic though, (was successful before, now fails with `no protocols available`):

```
ctx.ssl_version = "TLSv1"
ctx.min_version = OpenSSL::SSL::TLS1_VERSION
ctx.max_version = OpenSSL::SSL::TLS1_2_VERSION
```

src/test/ruby/ssl/test_socket.rb
MariuszCwikla: Ok, I've added integration test in `src/test/integration/ssl_test.rb` like you suggested.
`test_socket.rb` still needs improvement.

kares: should we remove this than, it's very dependend on what google.com has set on its servers, right?

src/main/java/org/jruby/ext/openssl/SSLContext.java
kares: right, yeah that one doing the `int` check before cast.

src/test/ruby/ssl/test_socket.rb
MariuszCwikla: Yes, let me remove it later since I've added the integration test. 
What do you think about the integration test then?

Later I will try to rework test in `test_socket.rb` or remove completely since it cannot be tested now due to another issue (`server` method fails with `no ciphers match`) at least on my env.

kares: > Yes, let me remove it later since I've added the integration test.

sorry but I would prefer not to deal with leftovers as they will bite someone later. 
I am fully aware CI testing is broken atm but that is no reason to leave externally dependent tests here.
you can keep it around for yourself in a follow-up branch.

> What do you think about the integration test then?

we do not run those atm regularly, so I am fine with having them as you did. thanks.

lib/jopenssl23/openssl/ssl.rb
kares: Thanks, will need to :eyes: at this since it changed behavior - which I did not expect to happen.

MariuszCwikla:
I had to rework this due to issue introduce by my previous commit. I hope now it's fine. It's bit greener now.
I've also added a parameterized test in `test_ssl.rb` to test matrix of cases.

This test also works fine with CRuby. Try this:

```
ruby src\test\ruby\ssl\test_ssl.rb -n "/test_ssl_minmax_.*/"
```

src/main/java/org/jruby/ext/openssl/SSLContext.java
MariuszCwikla: Done

src/test/ruby/ssl/test_socket.rb
MariuszCwikla: Done. I Removed this test.
Instead got inspired by `test_ssl.rb` and implemented parameterized test for this issue.

PR #216, opened 2020-11-04
Match OpenSSL::X509::Name.hash implementation with Ruby

### LINKED ISSUE #209: Jruby's OpenSSL::X509::Name.hash doesn't match ruby's
In the examples below, `cert` is globalsign's root-r1 ca cert.

On the command line I can output a cert's subject name hash with...
```
% openssl x509 -hash -noout < cert
5ad8a5d6
```
In ruby (eg 1.9 but also modern versions)
```
% ruby -e "require 'openssl'; puts (OpenSSL::X509::Certificate.new File.read('cert')).subject.hash.to_s(16)"
5ad8a5d6
```
In jruby
```
ruby -e "require 'openssl'; puts (OpenSSL::X509::Certificate.new File.read('cert')).subject.hash.to_s(16)"
-4f0c1892
```
It would appear that in jruby an old form of `hash` was implemented (correct for openssl  pre v1.0???) and never updated.
ruby's implementation has a `hash_old` method that kinda matches what jruby is still doing.

```
$ rvm use jruby
Using /home/fishera/.rvm/gems/jruby-9.2.9.0
$ ruby -e "require 'openssl'; puts [(OpenSSL::X509::Certificate.new File.read('cert')).subject.hash].pack('L').unpack('L').first.to_s(16)"
b0f3e76e
$ rvm use 1.9.3
Using /home/fishera/.rvm/gems/ruby-1.9.3-p551
$ ruby -e "require 'openssl'; puts (OpenSSL::X509::Certificate.new File.read('cert')).subject.hash_old.to_s(16)"
b0f3e76e
```
(Only kinda matches because I had to convert a signed int to unsigned as well before the output matched)

MariuszCwikla:
Fixes #209
Summary of changes:

* OpenSSL hash algorithm switched from MD5 to SHA1. Furthermore, for calculations they are using canonical encoding now (`X509_NAME_hash/asn1_string_canon`)
* MD5 version of `hash` is still present in Ruby API under `hash_old`
* `eql` and `<=>` are also affected by the change in OpenSSL (see `X509_NAME_cmp`)


kares:
Thanks Mariusz, well done!

PR #218, opened 2020-11-11
Fix integer overflow in hash_old

MariuszCwikla:
In #216 `X509Name.hash` was fixed and old implementation renamed to `hash_old`
However I discovered that old method still wasn't working correclty due to integer overflow:

> ruby -e "require 'openssl'; p OpenSSL::X509::Name.new([['CN', 'foo'], ['DC', 'bar']]).hash_old"
3294068023

> jruby -e "require 'openssl'; p OpenSSL::X509::Name.new([['CN', 'foo'], ['DC', 'bar']]).hash_old"
-1000899273

Ruby is using `unsinged int` for this calculation. In JRuby changing int to long solves the issue.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/X509Name.java
+++ b/src/main/java/org/jruby/ext/openssl/X509Name.java
@@ -633,7 +633,7 @@ public class X509Name extends RubyObject {
 
     @JRubyMethod
     public RubyFixnum hash_old() {
-        int hash;
+        long hash;
         try {
             hash = Name.hashOld( getX500Name() );
         }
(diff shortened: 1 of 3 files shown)


headius:
@MariuszCwikla The PR looks fine but the test should either be added to spec/ruby or test/jruby. The tests under test/mri are copied from CRuby and we do not edit them in our repo.

Thanks for the PR!

MariuszCwikla:
@headius thanks for hints but it seems that jruby-openssl is not consistent jruby repo.
MRI tests even for openssl module are in jruby repo (but disabled).
And there is no test/jruby under jruby-openssl, only test/ruby

Issue #249, opened 2022-02-10
revisit compatibility with gemified net/http


kares:
JRuby had a few modification to *net/http* in stdlib to work-around APIs JOSSL does not provide (e.g. `session_new_cb=`)
as well as some custom Java specific features (using the JVM system proxy).

ATM, once a gem-ifiied *net/http* is installed further `gem install` starts failing with the relevant errors e.g.
```
OpenSSL::SSL::SSLContext#session_cache_mode= has no effect under JRuby
ERROR:  While executing gem ... (NoMethodError)
    undefined method `session_new_cb=' for #<OpenSSL::SSL::SSLContext:0x1782896>
Did you mean?  session_id_context=
```

We could provide dummy APIs (such as `session_new_cb=`) even if they would do nothing as long as *net/http* does not rely on them as crucial and they do not decrease security.

Upstream case (for JRuby 9.4): https://github.com/jruby/jruby/pull/7088

PR #252, opened 2022-04-07
Add additional attr callbacks for session_cb

headius:
These are called by at least the default net-http gem, so we need
these methods to be present. I am not sure how much functionality
they need to have; in CRuby, they are just bound as attr readers/
writers.

See #249 and jruby/jruby#7088


kares:
most `_cb` we currently have actually work, these would be dummy and might cause confusion as they're never called.
believe a better approach would be to feature detect (on net-http): https://github.com/ruby/net-http/pull/52

headius:
The feature detect fix is better. We still need to revisit these callbacks but my PR is only partial and would silently pretend the callbacks are working.

PR #276, opened 2023-05-02
Fix for CRL verify when signed with EC key

### LINKED ISSUE #275: CRL check fails if the issuing CA uses EC key pair
The CRL file may be issued by a CA that uses EC key pair.  Validating peer certificates (of any key type) against CRL from this kind of CA currently fails.

The problem is that jruby-openssl currently cannot verify the CRL file if it is signed with EC public key.  The execution will end up in the default branch here, which fails when the key type is not RSA

https://github.com/jruby/jruby-openssl/blob/94aa9f466381f933078a014c3cf96675372836fa/src/main/java/org/jruby/ext/openssl/SecurityHelper.java#L613-L618

tsaarni:
This change adds support for verifying the signature CRL files that are signed using EC private key.

It allows using CRL files that are issued by CA that uses EC key pair. Previously the CRL check failed, and peer certificate was always rejected as invalid even if it was valid, since the validation of CRL file itself failed.

Fixes #275


tsaarni:
Hi @kares, 
I know that after last release there has not been that many changes yet, but since side-loading gems and versioning forks can get bit tricky, I thought I'd ask if you happen to have any plans for the next release (including this PR)?  🙏

PR #278, opened 2023-06-17
[deps] upgrade BC to latest 1.74

chadlwilson:
Bumps BouncyCastle to latest patched version.

Not sure if jruby-openssl uses the vulnerable LDAP stuff mentioned in [CVE-2023-33201](https://github.com/bcgit/bc-java/wiki/CVE-2023-33201) however this is probably good to reduce noise.

It also has the positive side effect of reducing uber-jar size, as "The now defunct PQC SIKE algorithm has been removed" which seems to reduce by ~3MB.

https://www.bouncycastle.org/releasenotes.html#r1rv74


kares:
Thanks Chad, will take a look - there's a couple of items that I will need to find free time to work on before pushing a release.

Reading the CVE I do not see JOSSL affected by this, hopefully I understood correctly and this can be tagged out for JRuby.

chadlwilson:
@kares Yeah, I don't think relevant to JRuby users either, but you know those noisy CVE scanners. 😅 Currently the CVE hasn't been disclosed via MITRE/NIST ([still RESERVED here](https://nvd.nist.gov/vuln/detail/CVE-2023-33201)) so it's not yet linked to bouncycastle CPE but when it is I guess folks may start complaining about transitive noise via jruby-complete and friends.

As a side note, I see there are no automated tests run against a JRuby 9.4.x release currently - not sure if that is of concern.

Issue #282, opened 2023-08-11
Calling `.to_der` raises " undefined method `[]' for #<OpenSSL::ASN1::Sequence>" error


skunkworker:
Recently I've been running into this error and was able to make a minimum reproducible example.

Expected (MRI openssl)

```ruby
data_sequence = ::OpenSSL::ASN1::Sequence([::OpenSSL::ASN1::Integer(0)])
asn1 = ::OpenSSL::ASN1::Sequence(data_sequence)
asn1.to_der =>  "0\x03\x02\x01\x00"
```

Actual (JRuby-openssl)

```ruby
data_sequence = ::OpenSSL::ASN1::Sequence([::OpenSSL::ASN1::Integer(0)])
asn1 = ::OpenSSL::ASN1::Sequence(data_sequence)
asn1.to_der => NoMethodError (undefined method `[]' for #<OpenSSL::ASN1::Sequence:0x389a9ff6>)
```

skunkworker:
Some of the backtrace.

```
 => #<OpenSSL::ASN1::Sequence:0xcb29b75 @value=[#<OpenSSL::ASN1::Integer:0x3674164b @value=0, @tag=2, @indefinite_length=false, @tagging=nil, @tag_class=:UNIVERSAL>], @tag=16, @indefinite_length=false, @tagging=nil, @tag_class=:UNIVERSAL>
jruby-9.3.9.0 :002 > asn1 = ::OpenSSL::ASN1::Sequence(data_sequence); asn1.to_der
2023-08-14T15:35:28.938-07:00 [main] INFO TraceType : Exception raised: NoMethodError : undefined method `[]' for #<OpenSSL::ASN1::Sequence:0xcb29b75>
2023-08-14T15:35:28.939-07:00 [main] INFO TraceType : Backtrace generated:
               getStackTrace at java/lang/Thread.java:1610
            getBacktraceData at org/jruby/runtime/backtrace/TraceType.java:247
                getBacktrace at org/jruby/runtime/backtrace/TraceType.java:53
            captureBacktrace at org/jruby/RubyException.java:402
                    preRaise at org/jruby/exceptions/RaiseException.java:208
                      <init> at org/jruby/exceptions/RaiseException.java:65
                      <init> at org/jruby/exceptions/Exception.java:39
                      <init> at org/jruby/exceptions/StandardError.java:38
                      <init> at org/jruby/exceptions/NameError.java:38
[...]

PR #283, opened 2023-08-15
[fix] try value#to_der when creating ASN.1 sequence

skunkworker:
Could fix https://github.com/jruby/jruby-openssl/issues/282

I noticed that there is an expectation that anything inside of an ASN1 object is either a RubyArray or a object that responds to `[]` this checks to see if the sub object responds to `#to_der` and calls that method. I was unable to get the test suite to run locally on my M1 pro, and would like to be able to test on here using the suites.

Works and doesn't work before this PR.

**Works**
```ruby
data_sequence = ::OpenSSL::ASN1::Sequence([::OpenSSL::ASN1::Integer(0)])
asn1 = ::OpenSSL::ASN1::Sequence([data_sequence])
asn1.to_der
```

**Doesn't work but should**
```ruby
data_sequence = ::OpenSSL::ASN1::Sequence([::OpenSSL::ASN1::Integer(0)])
asn1 = ::OpenSSL::ASN1::Sequence(data_sequence)
asn1.to_der
```

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/ASN1.java
+++ b/src/main/java/org/jruby/ext/openssl/ASN1.java
@@ -1935,6 +1935,10 @@ public class ASN1 {
                     if ( addEntry(context, vec, val.entry(i)) ) break;
                 }
             }
+            else if (respondsTo("to_der")) {
+                IRubyObject entry = value.callMethod(context, "to_der");
+                addEntry(context, vec, entry);
+            }
             else {
                 final int size = RubyInteger.num2int(value.callMethod(context, "size"));
                 for ( int i = 0; i < size; i++ ) {
--- a/src/test/ruby/test_asn1.rb
+++ b/src/test/ruby/test_asn1.rb
@@ -21,6 +21,18 @@ class TestASN1 < TestCase
     assert_equal i, OpenSSL::ASN1.decode(ai.to_der).value
   end
 
+  def test_encode_nested_sequence_to_der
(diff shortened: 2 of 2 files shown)


kares:
have rebased locally, the test unfortunately fail so seems like more investigation will be needed due :red_circle: 

```
Failure: test_encode_nested_sequence_to_der(TestASN1)
/oss/jruby-openssl/src/test/ruby/test_asn1.rb:27:in `test_encode_nested_sequence_to_der'
     24:   def test_encode_nested_sequence_to_der
     25:     data_sequence = ::OpenSSL::ASN1::Sequence([::OpenSSL::ASN1::Integer(0)])
     26:     asn1 = ::OpenSSL::ASN1::Sequence(data_sequence)
  => 27:     assert_equal "0\x03\x02\x01\x00", asn1.to_der
     28:   end
     29: 
     30:   def test_encode_nested_set_to_der
org/jruby/RubyKernel.java:1310:in `catch'
org/jruby/RubyKernel.java:1305:in `catch'
org/jruby/RubyKernel.java:1310:in `catch'
org/jruby/RubyKernel.java:1305:in `catch'
<"0\x03\x02\x01\x00">(US-ASCII) expected but was
<"0\x050\x03\x02\x01\x00">(ASCII-8BIT)


```

skunkworker:
> the original issue has been fixed, I have included the test case from the PR on master: [ceeacba](https://github.com/jruby/jruby-openssl/commit/ceeacbae1fa0a0b458c13968aed5bb0747a72f57)

Thanks for adding that, it is very appreciated.

PR #303, opened 2024-05-20
Making parse_config accepts a String like Ruby MRI

edipofederle:
With Ruby MRI, we can do:

`OpenSSL::Config.parse_config("") # also accepts a String` 

but not with JRuby:

```
NoMethodError: undefined method `lineno' for "":String
   Did you mean?  lines
```

### DIFF
--- a/lib/openssl/config.rb
+++ b/lib/openssl/config.rb
@@ -52,7 +52,11 @@ module OpenSSL
       # Raises a ConfigError on invalid configuration data.
       def parse_config(io)
         begin
-          parse_config_lines(io)
+          if(io.instance_of?(String))
+            parse(io)
+          else
+            parse_config_lines(io)
+          end
         rescue => error
           raise ConfigError, "error in line #{io.lineno}: " + error.message
         end


edipofederle:
@kares maybe we can take a quick look into this one? thanks

lib/openssl/config.rb
```diff
@@ -52,7 +52,11 @@ def parse(string)
       # Raises a ConfigError on invalid configuration data.
       def parse_config(io)
         begin
-          parse_config_lines(io)
+          if(io.instance_of?(String))
```
kares, line 55: I undestand the code is in native in MRI, but would be nice to follow existing Ruby conventions
```suggestion
          if io.instance_of?(String)
```

also could you please attempt to add test for this feature, Ruby seems to have parsing tests, so they could serve as a starting point: https://github.com/ruby/openssl/blob/master/test/openssl/test_config.rb

headius:
Seems like this is ready to go?

kares:
did look into this but did not figure out a path fwd. the logic isn't matching what MRI is doing (in native C)...
- `parse` isn't calling `parse_config` (we're a bit convoluted here `parse` -> `parse_config` -> `parse`)
- check is done on having an IO-like (gets ugly with JRuby) and everything else is considered `io.to_str`

@headius but if it looks good to you, feel free to merge. 
*I usually "Squash and merge" small change-sets (easier to dig out history) but if you have a different preference it's fine*

PR #304, opened 2024-05-20
Adding ConfigError and DEFAULT_CONFIG_FILE to .rb file

edipofederle:
Hi,

No sure if this is the right way, but since Config.java is deprecated, adding the ConfigError and the DEFAULT_CONFIG_FILE constant to ruby file.

### DIFF
--- a/lib/openssl/config.rb
+++ b/lib/openssl/config.rb
@@ -14,6 +14,8 @@
 require 'stringio'
 
 module OpenSSL
+  class ConfigError;end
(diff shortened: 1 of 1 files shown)


kares:
Thanks, there's some adjustments so I've added it separately: https://github.com/jruby/jruby-openssl/commit/5df8e30d641cbbedb8756242e6fd5c8b20500f51 (it's now available in 0.14.6)

Issue #308, opened 2024-06-11
OpenSSL::PKey::EC::Group does not support 4-argument constructor


headius:
The ...EC::Group class should support a four-element constructor, as seen in several tests for ...EC::Point: https://github.com/ruby/ruby/blob/7f407e0240ebd41d9fb1ea1bbd15442ed8744b34/test/openssl/test_pkey_ec.rb#L351

I've got a broken attempt to implement that form coming in a PR, but I'm unsure how to map the OpenSSL calls to BC yet.

PR #307, opened 2024-06-06
Implement Point#mul

headius:
Work in progress. Fixes #306.

PR #309, opened 2024-06-11
Implement 4-arg form of EC::Group.initialize

headius:
See #308.

This depends on the changes in #307.


kares:
shipped in 0.19.0

PR #325, opened 2025-01-03
Support more complex SANs in `verify_certificate_identity`

### LINKED ISSUE #324: `verify_certificate_identity` fails when `subjectAltName` includes `otherName`
`verify_certificate_identity` relies on a naive string-based approach to extract `DNS:` (and `IP:`) entries from the subjectAltName extension. However, the SAN is truly an ASN.1 structure, and when the library stringifies it, `otherName` entries and semicolons (or other delimiters) will appear in the output. This leads to the regex `split` call misreading the line, causing `verify_certificate_identity` to fail on certificates that include `otherName` (common in Active Directory environments, for example).

For instance, a SAN might be stringified like:

```
"subjectAltName = critical, otherName:[1.3.6.1.4.1.311.20.2.3, [CONTEXT 0]HOST1$@example.com];DNS:host1.example.com, DNS:example.com, DNS:MYDOMAIN"
```

(See https://github.com/jruby/jruby-openssl/blob/976a3f5152b36129ad478175473bd63345286450/src/main/java/org/jruby/ext/openssl/X509Extension.java#L518 for where this happens?)

When `split(/,\s+/)` is applied to this string (see https://github.com/jruby/jruby-openssl/blob/976a3f5152b36129ad478175473bd63345286450/lib/openssl/ssl.rb#L273C9-L273C32), the returned array is
```
[
  "subjectAltName = critical",
  "otherName:[1.3.6.1.4.1.311.20.2.3",
  "[CONTEXT 0]HOST1$@example.com];DNS:host1.example.com",
  "DNS:example.com", "DNS:MYDOMAIN"
]
```

The `DNS:host1.example.com` entry will not be found and extracted by the regex in `if /\ADNS:(.*)/` (https://github.com/jruby/jruby-openssl/blob/976a3f5152b36129ad478175473bd63345286450/lib/openssl/ssl.rb#L277C11-L277C26)

alexjfisher:
Not sure if this is the right approach, but I'll put it up for discussion anyway. :)

Fixes #324


lib/openssl/ssl.rb
```diff
@@ -270,26 +270,21 @@ def verify_certificate_identity(cert, hostname)
...
+        sequence = OpenSSL::ASN1.decode(ostr.value)
+        sequence.value.each{|san|
+          case san.tag
+          when 2 # dNSName in GeneralName (RFC5280)
             should_verify_common_name = false
-            return true if verify_hostname(hostname, $1)
-          # MRI 2.2.3 (JRuby parses ASN.1 differently)
-          #when 7 # iPAddress in GeneralName (RFC5280)
-          elsif /\AIP(?: Address)?:(.*)/ =~ general_name
+            return true if verify_hostname(hostname, san.value.last.value)
```
alexjfisher, line 279: In `ruby-openssl` this was `verify_hostname(hostname, san.value)`, but I've had to add `.last.value` to get this to work here, as I'm getting a single element Array returned instead of a plain string.  _Maybe_ this isn't the right solution though.  Should `san.value` instead return the exact same thing across both implementations??  I've not worked out where this is done though, and maybe it'd break loads of other things...

lib/openssl/ssl.rb
```diff
@@ -270,26 +270,21 @@ def verify_certificate_identity(cert, hostname)
...
-            return true if $1 == hostname
-            # NOTE: bellow logic makes little sense JRuby reads exts differently
-            # follows GENERAL_NAME_print() in x509v3/v3_alt.c
-            # if san.value.size == 4 || san.value.size == 16
-            #    begin
-            #      return true if $1 == IPAddr.new(hostname).to_s
-            #    rescue IPAddr::InvalidAddressError
-            #    end
-            # end
+            if san.value.last.value.size == 4 || san.value.last.value.size == 16
```
alexjfisher, line 282: Same here.  `san.value` is changed to `san.value.last.value`

src/test/ruby/ssl/fixtures/othername_cert/othername.cnf
```diff
@@ -0,0 +1,15 @@
+[ req ]
```
alexjfisher, line 1: I couldn't work out how to generate a test certificate with the `otherName` extension programatically, so used `openssl` on the command line with this config file.

```
openssl req -newkey rsa:2048 -nodes -x509 -days 36500 -keyout othername.key -out othername.crt -config othername.cnf -extensions req_ext
```

PR #329, opened 2025-03-13
Consider SHORTNAME and shortname equal as hostname

joerixaop:
There seems to be no reason to consider a local hostname as case sensitive when other hostnames are compared case insensitive.

If the behavior needs to be more strict it would be good to document in what way it is more strict than qualified domain names.


lib/openssl/ssl.rb
```diff
@@ -314,7 +314,7 @@ def verify_hostname(hostname, san) # :nodoc:
       san_parts = san.downcase.split(".")
 
       # TODO: this behavior should probably be more strict
-      return san == hostname if san_parts.size < 2
+      return san.downcase == hostname.downcase if san_parts.size < 2
```
kares, line 317: ssl.rb is mostly copied from openssl gem: https://github.com/ruby/openssl/blob/master/lib/openssl/ssl.rb#L317

you maybe want to post this upstream first, ideally with an actual test-case.

PR #330, opened 2025-04-17
Start adding support for missing features

segiddins:


headius:
We may want to use the same excludes logic as in JRuby's CI so we can track regressions. Thank you for kicking this off!

kares:
Some of this shipped in **0.19.0** (Ed25519), have added some more features based on the PR (e.g. X25519 support): https://github.com/jruby/jruby-openssl/commit/a7ca5a9fe9b4b58695d476dffd48f86b0ad4429e, https://github.com/jruby/jruby-openssl/commit/7eb532c5a0ec82f1be775dd1556c7beb3a61ce16 and https://github.com/jruby/jruby-openssl/commit/76ab17cd5b02ff77cc89400c0e1fdc08f42203ae these are expected to ship in next release.

Also we're now running some of the (relevant) sigstore tests in CI.

Thanks Samuel for starting the PR, we very much appreciate making JRuby not fall behind with latest of artifact security. :heart_decoration:

PR #334, opened 2025-06-29
Regenerate sample certificates to fix CI

smortex:
Some certificates used by CI have expired and cause CI failure.

Follow instructions from the `SETUP.txt` file to re-generate them, and update these instructions to make this easier the next time.


smortex:
Remaining failures might to be linked to jruby-head.

PR #335, opened 2025-06-30
Improving completeness of ASN1 encoding/decoding

HoneyryderChuck:
To atest some of the improvements of this PR, some encode/decode tests from the ruby-openssl repo were imported (which have been previously commented out).

In broad strokes, the main fixes/improvements from this changeset:

* Raw ASN1Data objects with array values can now be encoded/decoded.
* Improved support for indefinite length `Constructive` **and** raw `ASN1Data` objects (the checks around EndOfContent elements were ported from the `ruby-openssl` repo)
* Imported upstream check to avoid encoding indifinite length objects with a non-array value.
* initialization of `Primitive` objects was incorrect, it is now matching what upstream does
  * I contributed a rewrite in ruby of this logic to `ruby-openssl`, which `jruby-openssl` could also benefit of, but I don't think it's worth atm to just copy code around, more on that later.
* fixed internal `isEOC` check to take into account raw `ASN1Data` objects which "quack" like an `EndOfContent`
* `Primitive` objects with an OOTB-unsupported type will now raise an error **if** the value can't be encoded into a string (same as upstream).

I have more detailed information in each commit message, if you prefer that method of review. There were two tests I wasn't able to port, as making them work would require to adapt even more code from BC than I already did here, so I'll leave those for some other time.

On a separate note, it'd be really cool if [this](https://github.com/ruby/openssl/issues/20) could be taken into the finish line. It seems that it's hanging on licensing issues, and I don't know how hard it'd be to coalesce licenses, but besides the usual code copying churn and ability to track what's and what's not supported, I've been trying to lobby for more ruby in the source code ([here](https://github.com/ruby/openssl/pull/740) and [here](https://github.com/ruby/openssl/pull/777)), which would benefit both implementations, but it's hard to do it across multiple repos. 

cc @kares @headius

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/ASN1.java
+++ b/src/main/java/org/jruby/ext/openssl/ASN1.java
@@ -703,6 +703,10 @@ public class ASN1 {
         Constructive.addReadWriteAttribute(context, "tagging");
         Constructive.defineAnnotatedMethods(Constructive.class);
 
+        final ObjectAllocator eocAllocator = EndOfContent.ALLOCATOR;
+        RubyClass EndOfContent = ASN1.defineClassUnder("EndOfContent", _ASN1Data, eocAllocator);
+        EndOfContent.defineAnnotatedMethods(EndOfContent.class);
+
         ASN1.defineClassUnder("Boolean", Primitive, primitiveAllocator); // OpenSSL::ASN1::Boolean <=> value is a Boolean
         ASN1.defineClassUnder("Integer", Primitive, primitiveAllocator); // OpenSSL::ASN1::Integer <=> value is a Number
         ASN1.defineClassUnder("Null", Primitive, primitiveAllocator); // OpenSSL::ASN1::Null <=> value is always nil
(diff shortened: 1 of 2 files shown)


kares (commented):
Hey Tiago, thanks a lot - these changes look great from a first look :heart_decoration:. 

Will need to circle back for deeper review, for some of the parts, or at least resolve CI failures on master to make sure there aren't any regressions. Unless there are these changes should IMHO get into the next release.

HoneyryderChuck:
@kares thx for taking a look 🙏 I believe it's in a better state compatibility-wise (even if some of the java code is questionable, apologies for that 😂 ).

FYI I'm also working on the side to add the remainder of upstream asn1 tests here. It'll take way more time though, as openssl allows to build asn1 objects in too flexible a way when compared to BC (which just prevents certain combinations OTTB, and that's when one needs to import some of that code into here). But first things first, let's get this merged in once you give the sign-off away! 🙏

PR #338, opened 2025-12-10
Handle encrypted key pair exception

djberg96:
This is meant to address https://github.com/jruby/jruby/issues/8254 where an exception occurs if I try to generate a rubygem with a signed cert.

Note that this is AI generated code (I don't really know Java), so I'm happy to make changes if you think it needs them.


djberg96:
I've no idea why the one maven check is failing, help wanted.

kares (commented):
Thanks for the effort Danie, unfortunately this doesn't seem right.

Maybe we would be better of starting with a minimal reproducer (test-case).

PR #341, opened 2026-02-27
Fixes for JRuby 10.1 and Ruby 4.0 tests

headius:
A collection of updates and fixes to support JRuby 10.1 (Ruby 4.0 compatible).


kares:
not much left to do here, let's see how much excludes we have upon merging https://github.com/jruby/jruby/pull/9323 to master

PR #347, opened 2026-03-16
Fix SSL write data loss and flush-before-read on non-blocking writes

### LINKED ISSUE #242: gem push fails with Broken Pipe IOError
**Environment Information**
  
Provide at least:

**Expected Behavior**

**Actual Behavior**

```
    Broken pipe
```

jsvd:
Two bugs in SSLSocket cause data loss when using write_nonblock followed by sysread (the pattern used by net/http for POST requests):

1. write() unconditionally calls netWriteData.clear() after a non-blocking flushData(), discarding any encrypted bytes that weren't yet sent to the socket. Fix: use compact() which preserves unsent bytes.

2. sysreadImpl() never flushes pending netWriteData before reading. After the last write_nonblock, encrypted data can remain in the buffer and never reach the server. Fix: flush netWriteData at the start of sysreadImpl().

This was observed as EOFError during `gem push` on JRuby 10 with large gem files: the server never received the complete POST body and sent close_notify instead of an HTTP response.

fixes #242

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -688,7 +688,7 @@ public class SSLSocket extends RubyObject {
             if ( netWriteData.hasRemaining() ) {
                 flushData(blocking);
             }
-            netWriteData.clear();
+            netWriteData.compact();
             final SSLEngineResult result = engine.wrap(src, netWriteData);
             if ( result.getStatus() == SSLEngineResult.Status.CLOSED ) {
                 throw getRuntime().newIOError("closed SSL engine");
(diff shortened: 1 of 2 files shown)


src/main/java/org/jruby/ext/openssl/SSLSocket.java
```diff
@@ -830,6 +830,12 @@ private IRubyObject sysreadImpl(final ThreadContext context, final IRubyObject l
         }
 
         try {
+            // Flush any pending encrypted write data before reading, so the
+            // remote side receives the complete request before we wait for a response.
+            if ( engine != null && netWriteData.hasRemaining() ) {
+                flushData(true);
```
kares: btw. was this actually relevant in context of #242? 

also any thoughts about forcing a blocking `flushData` or the comment above in general (doing a write in context of user requesting a non-blocking read)?

kares (commented):
Noticed this late, seems like its just Opus wout a human touch?

What would be interesting to know is whether `gem push` got actually resolved after these or any other context you can share here related to the fix (for me the `gem push` issue never reproduced with Linux' defaults).

kares:
> I hoped there was enough of a human touch as I spend quite a bit of time experimenting and doing back and forths to arrive at this PR😅

Apologies, it's really hard to tell these days. :sweat_smile: 
It's all fine, except with non-trivial bugs making it explicit helps a lot. 

What you said in your last comment is really valuable to know. :heart_decoration:  

Appreciate the response, I'll take a look at the test once I am on #351..

PR #351, opened 2026-03-26
SSLSocket related improvements

### LINKED ISSUE #242: gem push fails with Broken Pipe IOError

### LINKED ISSUE #271: OpenSSL::Buffering#read_nonblock(..., exception: false) can still throw exceptions
Hello. I am not sure whether it's a jruby-openssl's or jruby's bug, but opening the issue here for now. Also, I don't have a way to reproduce, but it does reproduce with jruby-openssl 0.12.2 and 0.14.0 on jruby 9.2.21.0 and 9.3.9.0.

We're using httprb library, which has the following code (https://github.com/httprb/http/blob/main/lib/http/timeout/per_operation.rb#L41):
```
result = @socket.read_nonblock(size, buffer, :exception => false)
```

Which in my case leads to OpenSSL::Buffering#read_nonblock method. As you can see, it asks the method not to return the exception, but sometimes we get an exception from this method with class "OpenSSL::SSL::SSLErrorWaitReadable" and message "read would block" with this backtrace:

```
org/jruby/ext/openssl/SSLSocket.java:883:in `sysread_nonblock'
/opt/jruby/lib/ruby/stdlib/openssl/buffering.rb:205:in `read_nonblock'
/usr/local/bundle/gems/http-4.4.1/lib/http/timeout/per_operation.rb:63:in `block in readpartial'
org/jruby/RubyKernel.java:1442:in `loop'
/usr/local/bundle/gems/http-4.4.1/lib/http/timeout/per_operation.rb:62:in `readpartial'
```

As I said - I don't have neither a minimal code that reproduces the bug, nor an understanding on when it happens. For us it happened when we tried to connect to AWS-managed kubernetes apiserver (which means I can't get any info about what is happening there), but I assume that it should be pretty easy to understand where the problem lies for the person familiar with jruby-openssl and jruby itself.

cc @headius

### LINKED ISSUE #305: OpenSSL::SSL::SSLErrorWaitReadable with TLS 1.3
Starting in April we started seeing lots of "read would block" and OpenSSL::SSL::SSLErrorWaitReadable errors in our Sentry instance. The errors occur only with TLS 1.3 and not with TLS 1.2. This issue is hard to debug for us and we could not write a script to reproduce this error. The issue occurs for multiple different endpoints from different systems of different companies that we send HTTPS requests to, so it is definitely on our side.

We are using JRuby 9.3.14.0 and Java 11.0.22+0.

### LINKED ISSUE #317: "read would block" when communicating with external APIs
## Basic Info

* Faraday Version: `2.11.0`
* Ruby Version: `jruby 9.4.8.0 (3.1.4) 2024-07-02 4d41e55a67 OpenJDK 64-Bit Server VM 21.0.4+7-LTS on 21.0.4+7-LTS +jit [x86_64-linux]`
* JRuby Version: `0.15.0`

## Issue description
We occasionally and intermittently face `Faraday::SSLError`s with the message `read would block` when communicating with external endpoints. Our application is deployed as an ECS on EC2 service and (so far) we were unable to reproduce the issue locally or in our staging environment.

## Steps to reproduce
Unfortunately, we haven't managed to reproduce the issue locally. There are certain requests to external APIs that fail very consistently, however.

kares:


headius (commented):
Generally I approve of the fixes and they appear to make sense. The bulk of functional changes appear to be handling incomplete reads and writes better, rather than assuming completion and incorrectly dropping data.

The tests are harder to verify. They apear to be correct but there's a lot of duplicate boilerplate and they do not exist in any CRuby tests that I could find. There's some tests that are testing JRuby-specific internals via reflection and those tests may be problematic to maintain over time. They may also may fail as we further adopt JPMS and lose the ability to do arbitrary reflective access.

I'm not opposed to the tests in spirit but it's hundreds of lines of code to test largely the same server/client setup with slightly different behaviors. If these are valid test cases then they should probably be added to CRuby tests or ruby/spec and verified by other ruby/openssl committers who have knowledge of SSLSocket expected behaviors. I suspect some of these already exist in some form in the existing tests. They pass on CRuby (where not masked by a JRuby check) but what they are verifying is sometimes unclear to me.

PR #354, opened 2026-04-01
SSLSocket write and non-block fixes

### LINKED ISSUE #242: gem push fails with Broken Pipe IOError

### LINKED ISSUE #271: OpenSSL::Buffering#read_nonblock(..., exception: false) can still throw exceptions

kares:
another attempt at (*minimal*) fixing non-blocking write + read.

similar to #351 but wout test madness (tests here are still not great but we can revisit those later)

includes https://github.com/jruby/jruby-openssl/pull/347

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -141,6 +141,8 @@ public class SSLSocket extends RubyObject {
         return sites[ index.ordinal() ];
     }
 
+    private static final ByteBuffer EMPTY_DATA = ByteBuffer.allocate(0).asReadOnlyBuffer();
+
     SSLContext sslContext;
     private SSLEngine engine;
     private RubyIO io;
@@ -148,7 +150,6 @@ public class SSLSocket extends RubyObject {
     ByteBuffer appReadData;
     ByteBuffer netReadData;
     ByteBuffer netWriteData;
-    private final ByteBuffer dummy = ByteBuffer.allocate(0); // could be static
 
     private boolean initialHandshake = false;
     private transient long initializeTime;
@@ -610,7 +611,7 @@ public class SSLSocket extends RubyObject {
 
     private void doWrap(final boolean blocking) throws IOException {
         netWriteData.clear();
-        SSLEngineResult result = engine.wrap(dummy, netWriteData);
+        SSLEngineResult result = engine.wrap(EMPTY_DATA.duplicate(), netWriteData);
         netWriteData.flip();
         handshakeStatus = result.getHandshakeStatus();
         status = result.getStatus();
@@ -794,7 +795,7 @@ public class SSLSocket extends RubyObject {
         }
         netWriteData.clear();
         try {
-            engine.wrap(dummy, netWriteData); // send close (after sslEngine.closeOutbound)
+            engine.wrap(EMPTY_DATA.duplicate(), netWriteData); // send close (after sslEngine.closeOutbound)
         }
         catch (SSLException e) {
             debug(getRuntime(), "SSLSocket.doShutdown", e);
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -145,9 +145,9 @@ public class SSLSocket extends RubyObject {
     private SSLEngine engine;
     private RubyIO io;
 
-    private ByteBuffer appReadData;
-    private ByteBuffer netReadData;
-    private ByteBuffer netWriteData;
+    ByteBuffer appReadData;
+    ByteBuffer netReadData;
+    ByteBuffer netWriteData;
     private final ByteBuffer dummy = ByteBuffer.allocate(0); // could be static
 
     private boolean initialHandshake = false;
--- a/src/main/java/org/jruby/ext/openssl/SSLSocket.java
+++ b/src/main/java/org/jruby/ext/openssl/SSLSocket.java
@@ -553,10 +553,6 @@ public class SSLSocket extends RubyObject {
         result[0] = WRITE_WOULD_BLOCK_RESULT;
     }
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/jruby/ext/openssl/SSLSocket.java
```diff
@@ -614,7 +611,7 @@ private IRubyObject doHandshake(final boolean blocking, final boolean exception)
 
     private void doWrap(final boolean blocking) throws IOException {
         netWriteData.clear();
-        SSLEngineResult result = engine.wrap(dummy, netWriteData);
+        SSLEngineResult result = engine.wrap(EMPTY_DATA.duplicate(), netWriteData);
```
headius, line 614: Is it necessary to duplicate EMPTY_DATA? It is already zero-length and marked as read-only.

src/main/java/org/jruby/ext/openssl/SSLSocket.java
```diff
@@ -793,7 +795,7 @@ private void doShutdown() throws IOException {
         }
         netWriteData.clear();
         try {
-            engine.wrap(dummy, netWriteData); // send close (after sslEngine.closeOutbound)
+            engine.wrap(EMPTY_DATA.duplicate(), netWriteData); // send close (after sslEngine.closeOutbound)
```
headius, line 798: Ditto here, why duplicate?

kares:
@jsvd have included #347 here, if you have any opinions please share

src/test/java/org/jruby/ext/openssl/SSLSocketTest.java
```diff
@@ -0,0 +1,251 @@
...
+            RubyString payload = RubyString.newString(runtime, chunk);
+
+            long totalSent = 0;
+            for (int i = 0; i < 64; i++) { // 64 * 4KB = 256KB
+                try {
+                    IRubyObject written = client.syswrite_nonblock(currentContext(), payload);
+                    totalSent += ((RubyInteger) written).getLongValue();
+                } catch (RaiseException e) {
+                    if ("OpenSSL::SSL::SSLErrorWaitWritable".equals(e.getException().getMetaClass().getName())) {
+                        System.out.println("syswrite_nonblock expected: " + e.getMessage());
```
headius, line 77: Shouldn't these be asserts or fails, so they show up as the cause of failure? stdout printlns can get lost in test harnesses.

src/test/java/org/jruby/ext/openssl/SSLSocketTest.java
```diff
@@ -0,0 +1,251 @@
...
+            try {
+                RubyFixnum len = RubyFixnum.newFixnum(runtime, 8192);
+                while (true) {
+                    IRubyObject data = server.sysread(currentContext(), len);
+                    serverReceived[0] += ((RubyString) data).getByteList().getRealSize();
+                }
+            } catch (RaiseException e) {
+                String errorName = e.getException().getMetaClass().getName();
+                if ("EOFError".equals(errorName) || "IOError".equals(errorName)) { // client closes connection
+                    System.out.println("server-reader expected: " + e.getMessage());
```
headius, line 113: Again seems weird to have printlns in a test, even if it's running in a separate thread.

src/test/java/org/jruby/ext/openssl/SSLSocketTest.java
```diff
@@ -0,0 +1,251 @@
...
+                } else {
+                    System.err.println("server-reader unexpected: " + e.getMessage());
```
headius, line 115: Now we're outputting to err?

headius (commented):
The batch of fixes looks pretty good after a first review pass. The tests are heinous but you know that. I suppose in the interest of getting fixes landed we can go with them for now but they shouldn't stay like this for long. It's very hard to follow what they're testing, or if they're actually testing what they should be.

src/main/java/org/jruby/ext/openssl/SSLSocket.java
kares: shouldn't be, except `engine.wrap` in theory does set `limit` although likely not on an empty buffer.

src/test/java/org/jruby/ext/openssl/SSLSocketTest.java
kares: this part is mostly `debug` messages, the other branch fails.

they're useful esp. since a lot of time these exceptions do not have a trace.

PR #356, opened 2026-04-10
Fix X509::Certificate#dup dropping all extensions

sferik:
`OpenSSL::X509::Certificate#dup` produces an invalid copy. The `initialize_copy` method checked for identity and frozenness but never copied any certificate state from the source object, so the duplicate could lose its extensions, subject, issuer, serial, etc.

As far as I can tell, this has been latent for a long time but was surfaced following the addition of `Certificate#tbs_bytes` in https://github.com/jruby/jruby-openssl/commit/ee7f0c87c79c36ca80f19090589ad0f3abcea45e. That addition caused `sigstore-ruby` to enter a code path that dups a Fulcio-issued certificate and iterates its extensions, which were silently empty, causing `gem push` with Sigstore attestation to fail on JRuby:

```
No PrecertificateSignedCertificateTimestamps found for the certificate (Sigstore::Error::InvalidCertificate)
```

Here is an example of a Java-platform gem push failing in CI: https://github.com/sferik/multi_json/actions/runs/24221466355/job/70713680096

I was able to successfully work around this issue by [using CRuby for the signing step](https://github.com/sferik/multi_json/commit/f262be6df0427904ebab51b27618ffbff47a7052).

Here is a minimal reproduction of this issue:

```ruby
require "openssl"

key = OpenSSL::PKey::RSA.new(2048)
cert = OpenSSL::X509::Certificate.new
cert.version = 2
cert.serial = 1
cert.subject = OpenSSL::X509::Name.new([["CN", "test"]])
cert.issuer = cert.subject
cert.not_before = Time.now
cert.not_after = Time.now + 3600
cert.public_key = key.public_key

ef = OpenSSL::X509::ExtensionFactory.new
ef.subject_certificate = cert
ef.issuer_certificate = cert
cert.extensions = [ef.create_extension("subjectKeyIdentifier", "hash")]
cert.sign(key, OpenSSL::Digest::SHA256.new)

duped = cert.dup

puts "Original extensions: #{cert.extensions.map(&:oid)}"
puts "Duped extensions:    #{duped.extensions.map(&:oid)}"
```

Expected: both lines print `["subjectKeyIdentifier"]`
Actual: duped prints `[]`

This patch fixes `Certificate#dup` by copying the live certificate state instead of leaving the duplicate empty.

For signed certificates, it clones the underlying parsed certificate and also copies the current Ruby-side fields.

For unsigned certificates still under construction, it deep-copies mutable fields.

This patch also marks `extensions=` as a mutating operation so the object’s state stays consistent after replacing extensions on an already-signed certificate.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/X509Cert.java
+++ b/src/main/java/org/jruby/ext/openssl/X509Cert.java
@@ -292,7 +292,51 @@ public class X509Cert extends RubyObject {
         if ( this == obj ) return this;
 
         checkFrozen();
+        super.initialize_copy(obj);
+
+        copyState(getRuntime().getCurrentContext(), (X509Cert) obj);
         return this;
+    }
+
+    private void copyState(final ThreadContext context, final X509Cert that) {
+        final Ruby runtime = context.runtime;
+
+        this.subject = copyName(context, that.subject);
+        this.issuer = copyName(context, that.issuer);
+        this.serial = that.serial;
+        this.not_before = copyTime(runtime, that.not_before);
+        this.not_after = copyTime(runtime, that.not_after);
+        this.sig_alg = that.sig_alg == null ? null : that.sig_alg.dup();
+        this.version = that.version;
+        this.cert = copyCertificate(context, that.cert);
+        this.public_key = that.public_key == null ? null : (PKey) that.public_key.dup();
+
+        this.extensions.clear();
+        for ( X509Extension ext : that.extensions ) {
+            this.extensions.add( (X509Extension) ext.dup() );
+        }
+
+        this.changed = that.changed;
+    }
+
+    private static IRubyObject copyName(final ThreadContext context, final IRubyObject name) {
+        if ( name == null || name.isNil() ) return name;
+        return X509Name.newName(context.runtime, ((X509Name) name).getX500Name());
+    }
+
+    private static RubyTime copyTime(final Ruby runtime, final RubyTime time) {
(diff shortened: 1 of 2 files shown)


headius:
This is great background, thank you! I'll review this along with @kares and @enebo and we'll get it into the next release.

sferik:
@headius Thanks! I must admit, it’s been a while since I’ve written Java, so please treat this PR as a bug report that happens to have some code attached. I’m not in any way attached to my proposed solution, just as long as the problem gets solved, one way or another.

src/main/java/org/jruby/ext/openssl/X509Cert.java
```diff
@@ -292,9 +292,53 @@ public IRubyObject initialize_copy(IRubyObject obj) {
         if ( this == obj ) return this;
 
         checkFrozen();
+        super.initialize_copy(obj);
+
+        copyState(getRuntime().getCurrentContext(), (X509Cert) obj);
         return this;
     }
 
+    private void copyState(final ThreadContext context, final X509Cert that) {
```
kares: you don't really use the `context`, so just pass around the `runtime`.

the `this`/`that` naming is different (confusing) from the usual Java semantics (e.g. `equals`).

you probably want to use different naming of simply inline this particular method.

sferik: Resolved in https://github.com/jruby/jruby-openssl/pull/356/changes/26158ef52b5adaf6bcda420011a7425ace1557bc. Please let me know if you have any other feedback and if you want me to squash these commits into one.

PR #360, opened 2026-04-20
[deps] upgrade BC to version 1.84

chadlwilson:
Replaces #357 and #359

https://www.bouncycastle.org/download/bouncy-castle-java/?filter=java%3Drelease-1-84

Fixes a number of CVEs which may or may not affect JRuby, but cause scanner noise all the same.
- CVE-2025-14813 - GOSTCTR implementation unable to process more than 255 blocks correctly.
- CVE-2026-0636 - LDAP Injection Vulnerability in LDAPStoreHelper.java.
- CVE-2026-3505 - Unbounded PGP AEAD chunk size leads to pre-auth resource exhaustion.
- CVE-2026-5588 - PKIX draft CompositeVerifier accepts empty signature sequence as valid.
- CVE-2026-5598 - Non-constant time comparisons risk private key leakage in FrodoKEM.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -34,7 +34,7 @@ Rake::TestTask.new do |task|
   test_files = FileList['src/test/ruby/**/test*.rb'].to_a
   task.test_files = test_files.map { |path| path.sub('src/test/ruby/', '') }
   task.verbose = true
-  task.loader = :direct
+  task.loader = "ARGV.each { |f| require f unless f.start_with?('-') }"
   task.ruby_opts = [ '-C', 'src/test/ruby', '-rbundler/setup' ]
 end
 task :test => 'lib/jopenssl.jar'
@@ -49,7 +49,7 @@ namespace :integration do
     unless File.exist?(File.join(it_path, 'Gemfile.lock'))
       raise "bundle not installed, run `rake integration:install'"
     end
-    loader = "ARGV.each { |f| require f }"
+    loader = "ARGV.each { |f| require f unless f.start_with?('-') }"
     lib = [ File.expand_path('../lib', __FILE__), it_path ]
     test_files = FileList['src/test/integration/*_test.rb'].map { |path| path.sub('src/test/integration/', '') }
     ruby "-I#{lib.join(':')} -C src/test/integration -e \"#{loader}\" #{test_files.map { |f| "\"#{f}\"" }.join(' ')}"
--- a/src/main/java/org/jruby/ext/openssl/impl/PKey.java
+++ b/src/main/java/org/jruby/ext/openssl/impl/PKey.java
@@ -297,7 +297,7 @@ public class PKey {
             org.bouncycastle.asn1.sec.ECPrivateKey key = org.bouncycastle.asn1.sec.ECPrivateKey.getInstance(seq);
             AlgorithmIdentifier algId = keyInfo.getPrivateKeyAlgorithm();
             if (algId == null) { // mockPrivateKeyInfo
-                algId = new AlgorithmIdentifier(X9ObjectIdentifiers.id_ecPublicKey, key.getParameters());
+                algId = new AlgorithmIdentifier(X9ObjectIdentifiers.id_ecPublicKey, key.getParametersObject().toASN1Primitive());
             }
             final PrivateKeyInfo privInfo = new PrivateKeyInfo(algId, key);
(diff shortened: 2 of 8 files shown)


src/main/java/org/jruby/ext/openssl/impl/PKey.java
```diff
@@ -297,7 +297,7 @@ public static KeyPair readECPrivateKey(final KeyFactory keyFactory, final Privat
             org.bouncycastle.asn1.sec.ECPrivateKey key = org.bouncycastle.asn1.sec.ECPrivateKey.getInstance(seq);
             AlgorithmIdentifier algId = keyInfo.getPrivateKeyAlgorithm();
             if (algId == null) { // mockPrivateKeyInfo
-                algId = new AlgorithmIdentifier(X9ObjectIdentifiers.id_ecPublicKey, key.getParameters());
+                algId = new AlgorithmIdentifier(X9ObjectIdentifiers.id_ecPublicKey, key.getParametersObject().toASN1Primitive());
```
chadlwilson, line 300: Deprecated method removed; inlined to the deprecated implementation. Replacement has been there at least since `1.78` so should be sufficiently backwards compatible.

https://github.com/bcgit/bc-java/blob/30c6cc60ef5aa9062a083a8cea3e5c4f96d91a2a/core/src/main/java/org/bouncycastle/asn1/sec/ECPrivateKey.java#L147-L154

Rakefile
```diff
@@ -49,7 +49,7 @@ namespace :integration do
     unless File.exist?(File.join(it_path, 'Gemfile.lock'))
       raise "bundle not installed, run `rake integration:install'"
     end
-    loader = "ARGV.each { |f| require f }"
+    loader = "ARGV.each { |f| require f unless f.start_with?('-') }"
     lib = [ File.expand_path('../lib', __FILE__), it_path ]
```
chadlwilson, line 53: Rake 13.4.x change semantics of verbose flag which broke `:direct` so working around this per https://github.com/ruby/rake/issues/724

kares: oh thanks, was very confused about this one.

src/main/java/org/jruby/ext/openssl/impl/PKey.java
kares: recall there was a reason it's been kept, not sure. 
think we might push master as 0.16.x anyway (due other updates).

chadlwilson: Is master able to be used on 9.4 without other changes to JRuby itself? Preferably we could preserve ability to somehow release a version suitable for 9.4 due to the CVE noise even if we create a branch off a previous 0.15.x? (that seems like extra hassle though)

kares: > Is master able to be used on 9.4 without other changes to JRuby itself?

current compat is (still) >= 9.2 ish, although we should be dropping < 9.3 at least maybe even 9.3 soon

chadlwilson: I saw that, but the test setup is rather complex so I couldn't quite figure out if the tests were in fact running against all those versions on GHA or not. Didn't spend much time though.

In any case, I guess that means we should be able to upgrade jruby 9.4's bundled version from `0.15.5` to `0.15.8|0.16.0` as well. I presume upgrading the stdlib bundled version is a bit higher risk since it's not user controlled; but this is what is required to get scanners to stop complaining - since they see the BC jars inside and 9.4 has some ancient ones even if you add/override the gem yourself from outside, e.g

```
One or more dependencies were identified with known vulnerabilities in gocd:

jruby-stdlib-9.4.14.0.jar: bcpkix-jdk18on-1.79.jar (pkg:maven/org.bouncycastle/bcpkix-jdk18on@1.79, cpe:2.3:a:bouncycastle:bouncy_castle_for_java:1.79:*:*:*:*:*:*:*) : CVE-2026-5588
jruby-stdlib-9.4.14.0.jar: bcprov-jdk18on-1.79.jar (pkg:maven/org.bouncycastle/bcprov-jdk18on@1.79, cpe:2.3:a:bouncycastle:bouncy_castle_for_java:1.79:*:*:*:*:*:*:*) : CVE-2026-5598, CVE-2026-5588, CVE-2026-0636
```

Rakefile
kares: heh, this ended up only running a few of the .rb tests not the whole suite. :sweat_smile:

PR #363, opened 2026-06-21
Fix X509::Certificate#tbs_bytes to honor extensions= mutations

segiddins:
## Problem

`OpenSSL::X509::Certificate#tbs_bytes` re-encoded the certificate's *original* cached `TBSCertificate` and ignored any prior in-place mutation (e.g. `cert.extensions=`). MRI's `openssl` re-encodes from the certificate's current state (`i2d_re_X509_tbs`), so removing an extension and then calling `tbs_bytes` reflects the removal.

This divergence breaks Certificate Transparency / SCT verification, which reconstructs a precertificate TBS by cloning the leaf certificate, deleting the `ct_precert_scts` extension, and signing/verifying over `tbs_bytes`. Surfaced while working on sigstore/sigstore-ruby#308.

```ruby
cert = OpenSSL::X509::Certificate.new(File.read("cert-with-scts.pem"))
dup  = cert.dup
dup.extensions = dup.extensions.reject { |e| e.oid == "ct_precert_scts" }
dup.extensions.map(&:oid)   # SCT extension gone from the Ruby view
dup.tbs_bytes.bytesize      # before: unchanged (stale); after: reflects the removal
```

## Fix

- `tbs_bytes` now re-encodes the `TBSCertificate` from the certificate's current state when it has been mutated (preserving the non-extension fields verbatim and taking the live extension set), and returns the cached TBS otherwise. This keeps it consistent with `extensions=` and matches MRI's semantics.
- Extensions are now parsed in DER (on-the-wire) order rather than critical-then-non-critical set order. This matches MRI's `X509#extensions` and is required for the rebuilt precertificate TBS to carry extensions in the order the log signed; without it, SCT verification would still fail.

Verified byte-identical to MRI (OpenSSL 3.5.1) for the unmutated case and for single-extension removal. The full `x509` test suite passes.

## Notes

- Like the existing `to_der`, non-extension setters (`serial=`, `subject=`, ...) are reflected only after a subsequent `sign`; `tbs_bytes` reflects the extension set, which is what CT precertificate reconstruction needs.
- Removing *all* extensions yields valid RFC 5280 DER (the field is omitted), whereas MRI emits an empty extensions field; this degenerate case is outside the CT scenario.
- Adds specs mirroring MRI coverage: `tbs_bytes` equals the TBS slice of `to_der` when unmutated, preserves DER extension order, reflects `extensions=` removal, and is unchanged by a no-op reassignment.

Note: stable (10.0.2.0) lacks `tbs_bytes` entirely; this fixes the broken implementation present on the Ruby-4.0-compat line.

### DIFF
--- a/src/main/java/org/jruby/ext/openssl/X509Cert.java
+++ b/src/main/java/org/jruby/ext/openssl/X509Cert.java
@@ -53,15 +53,22 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import org.bouncycastle.asn1.ASN1BitString;
 import org.bouncycastle.asn1.ASN1Encodable;
 import org.bouncycastle.asn1.ASN1EncodableVector;
+import org.bouncycastle.asn1.ASN1Encoding;
 import org.bouncycastle.asn1.ASN1ObjectIdentifier;
 import org.bouncycastle.asn1.ASN1Sequence;
+import org.bouncycastle.asn1.DERBitString;
 import org.bouncycastle.asn1.DLSequence;
+import org.bouncycastle.asn1.x509.Extensions;
+import org.bouncycastle.asn1.x509.ExtensionsGenerator;
 import org.bouncycastle.asn1.x509.GeneralName;
 import org.bouncycastle.asn1.x509.GeneralNames;
 
 import org.bouncycastle.asn1.x509.SubjectPublicKeyInfo;
+import org.bouncycastle.asn1.x509.TBSCertificate;
+import org.bouncycastle.asn1.x509.V3TBSCertificateGenerator;
 import org.bouncycastle.cert.X509CertificateHolder;
 import org.bouncycastle.cert.X509v3CertificateBuilder;
 import org.bouncycastle.operator.ContentSigner;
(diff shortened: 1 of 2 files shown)


headius:
Explanation makes sense and I gather that this issue was always there but did not really manifest until `tbs_bytes` came along. Is that correct?

Also,

> Note: stable (10.0.2.0) lacks tbs_bytes entirely; this fixes the broken implementation present on the Ruby-4.0-compat line.

Does this mean that only the 4.0 line is affected?

headius:
Note to self and @kares: need to get 10.1 in here.

rake test failures might be due to old test sources that expect the original cert inspect incorrectly?

kares:
PR seems fully LLM generated, would be nice to have less of the chit-chat (shorted human reviewed explanations) in comments at least. 

The direct BC approach is a bit of a complication, for some of the work towards FIPS, would be nice to try out some alternatives, if time allows.

The actual bug seems to be clear from the description, so that's :+1:

kares:
Appreciate the PR, but this seems like there's a lot going on; to be fair, the starting point wasn't super clear.

Have setup an alternative: https://github.com/jruby/jruby-openssl/pull/364 copied the tests from here and they pass.

segiddins:
Appreciate you taking this over @kares !

kares:
merged https://github.com/jruby/jruby-openssl/pull/364

PR #365, opened 2026-07-10
Load `openssl` when BouncyCastle FIPS jars are on the classpath

tehbooom:
## Summary

`require "openssl"` crashes when BouncyCastle FIPS jars (bc-fips, bcpkix-fips, bctls-fips, bcutil-fips) are on the classpath in place of non-FIPS BouncyCastle. Several classes reference bc-internal types under `org.bouncycastle.jce.provider.*` and `org.bouncycastle.jcajce.provider.asymmetric.util.*` that don't exist in the FIPS distribution, so the class-load hits `NoClassDefFoundError` before `openssl` can initialize. 

This PR makes jruby-openssl load under either BouncyCastle distribution. It is a compatibility fix, not a feature. It does not add FIPS enforcement and does not change which provider is selected.

## Approach

Every code path that references a bc-internal type absent from the FIPS jars is moved verbatim into a lazily-loaded `BCInternal` nested class, and the outer method delegates to it. Because inner classes compile to separate class files and the verifier resolves referenced types per-class, the outer classes no longer carry any load-time reference to the absent types, so the class loads regardless of which BC distribution is present. Code paths that genuinely require the absent classes fail at call time (only if invoked), not at load time.

The non-FIPS path is unchanged. The moved method bodies are byte-identical, and delegation only changes where the code lives, not what it does.

## Changes

Isolation into `BCInternal` nested classes (verbatim extraction, delegated): `SecurityHelper`, `PKeyEC`, `PKeyRSA`, `PKCS5`, `X509Name`, `PEMInputOutput`, `x509store/Name`, `x509store/X509AuxCertificate`, `impl/PKey`.

DN comparison: removes the `X509CertificateObject` instanceof fast-path (that class is bc-internal and absent from FIPS jars) and uses `X500Principal` comparison, which is RFC-4514 normalized and encoding-tolerant (verified to treat PrintableString and UTF8String of the same subject as equal).

EC `dsa_verify` error message: a non-sequence signature previously produced a double-prefixed message (`invalid signature: invalid signature (not a sequence)`); this restores the single, correct message.

## Behavior-change note

Removing the `X509CertificateObject` DN fast-path is correctness-preserving but a minor performance regression for non-FIPS users doing heavy certificate-chain and CRL verification, where the old O(1) fast-path is no longer taken. The fast-path was already unavailable under FIPS (the class is absent), so this affects only the non-FIPS path, and only performance, not results.

## Testing

Full existing suite passes unchanged (no behavior change on the non-FIPS path).

New regression test for the EC `dsa_verify` message, verified to fail on the pre-fix double-prefix behavior and pass on the fix.

New test confirming DN comparison matches subjects encoded as PrintableString versus UTF8String.

Verified `require "openssl"` loads clean with only bc-fips jars on the classpath (non-FIPS BC absent), and that operations depending on absent bc-internal classes fail at call time rather than at load time.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -281,6 +281,7 @@ DO NOT MODIFY - GENERATED CODE
         <configuration>
           <source>1.8</source>
           <target>1.8</target>
+          <release>8</release>
           <encoding>UTF-8</encoding>
           <debug>true</debug>
           <showWarnings>true</showWarnings>
--- a/src/main/java/org/jruby/ext/openssl/PKCS5.java
+++ b/src/main/java/org/jruby/ext/openssl/PKCS5.java
@@ -27,10 +27,6 @@ import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 import javax.crypto.Mac;
 
-import org.bouncycastle.crypto.CipherParameters;
-import org.bouncycastle.crypto.PBEParametersGenerator;
-import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
-import org.bouncycastle.crypto.params.KeyParameter;
 
 import org.jruby.Ruby;
 import org.jruby.RubyModule;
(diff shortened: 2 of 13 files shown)


tehbooom:
Insufficient and will not work. Closing.

kares:
Yep, it's not trivial working around having both BCs on class-path.

We're planning a separate (jruby-openssl-fips) release artifact...

headius:
FWIW we do load BC into a JRuby-specific classloader, but if there's other BC jars in classloaders above JRuby this problem is very difficult to avoid.

@tehbooom Get in touch with @kares if you are interested in FIPS support, which would largely eliminate this issue.

Or get in touch with me directly; I can help support your JRuby use and act on your behalf wrt getting access to FIPS support (by working with @kares). https://headius.com/services

PR #367, opened 2026-07-17
[deps] upgrade BC to version 1.85

Sharpie:
BouncyCastle 1.85 was tagged at the beginning of the week. As of today, Friday July 17th, the release notes have not been published to the website:

https://www.bouncycastle.org/download/bouncy-castle-java/#release-notes

However, the git docs indicate nearly _three dozen_ CVE fixes in this release:

https://github.com/bcgit/bc-java/blob/r1rv85/docs/releasenotes.html#L332-L363

This PR bumps the version included with `jruby-openssl` to 1.85.


Sharpie:
One item to be aware of: BC 1.85 started enforcing the 64 character limit on certificate `CN=` fields that is specified by RFC 5280. This makes the new library version unable to manipulate long certificate names that were previously fine.

They did add an example of how to work around the new behavior:

https://github.com/bcgit/bc-java/commit/a6ae24643e7ba7cfdf9267656ce83bb656ab9777

Sharpie:
And a quick follow-up: it looks like the length enforcement in BouncyCastle 1.85 doesn't affect the `jruby-openssl` implementation of things like `OpenSSL::X509::Name` because Strings are pre-encoded to DER format and the character limit is only enforced when BouncyCastle processes a String value.

kares:
Thanks, I've run the check with a bunch of external test suites as well and everything seemed :green_circle:

PR #375, opened 2026-09-21
[deps] bump Bouncy Castle to 1.86

chadlwilson:
Many more CVE fixes here: https://www.bouncycastle.org/resources/new-release-bouncy-castle-java-1-86/

Seems we need a 0.16.x branch, as we need a way to get security patches out into released jruby versions and master is already on `0.19.x`?

IMHO there is a truly excessive amount of boilerplate here just to add a new version to a matrix...


headius:
I'll look into branch.

chadlwilson:
The test matrix also seems totally out of control to me, which makes this take absolutely forever (and quite the waste of energy)
- testing on non-LTS Java 24?
- testing with multiple different patch versions of each of 9.2, 9.3, 9.4, 10 etc? Why? Surely we don't "support" an older jruby patch version once a new patch release comes out?
- still testing with JRuby 9.2? that's not in the declared supported range for `0.16.x`?
- still testing with JRuby 9.3? Very, very EOL? 

Anyway, seems there are some other problems here, so needs work.

headius:
I'm going to roll with JRuby releases with just latest 0.16.x for now since these are fairly recent (no CVSS and some don't show up in databases yet) and this needs more time to clean up.

@kares We'll want to spin a 0.16.x with this update too, but I'm not sure where to branch from.

headius:
@chadlwilson The matrix could definitely be pared down. No non-LTS Java (except perhaps most recent), no JRuby releases prior to 9.3 or 9.4 (10 would make it impossible to do any extended 9.4 support), etc.

chadlwilson:
> I'm going to roll with JRuby releases with just latest 0.16.x for now since these are fairly recent (no CVSS and some don't show up in databases yet) and this needs more time to clean up.

Yeah, that makes sense.

I haven't actually analyzed whether the CVEs are relevant to jruby-openssl usage, but I'd suggest that in a world of AI, we're gonna need to find a way to streamline the process for getting updates through the process into a jruby release, particularly for vendored gems like jruby-openssl, since they make breaking changes alongside non-backported security fixes. (presumably unless you're a paying customer)

headius:
When we can loop in @kares we'll have a chat about long term management of jruby-openssl and BC.

I have pushed a "0.16" branch based on the "v0.16.2" tag.

chadlwilson:
Fixed now. There was a patch released to `bctls` which is needed.

We should probably migrate this to use the BOM to define dependencies rather than managing individual artifact versions to avoid issues like this: https://mvnrepository.com/artifact/org.bouncycastle/bc-jdk18on-bom

kares:
I've started updating the 0.16 branch getting it somehow into shape for another release. This MR as is, isn't really useful in that direction.

Going to pass on the "out-of-control" claims, there's a (mostly security) reasoning but I don't think this needs to be discussed here.

chadlwilson:
Yeah, once this is merged, I can open another PR to cherry-pick `0.16`. Or alternatively I can re-target this one for `0.16.x` and we can merge forward, if that is the branch mgmt plan?

chadlwilson:
`master` seems in a broken state after https://github.com/jruby/jruby-openssl/commit/3f5514b289c97ec02ef11f19aa82a5179bf60dd8, so closing this.

PR #377, opened 2026-09-22
[test] correct CRL test to be agnostic of local TZ

chadlwilson:
This test fails if running in a local TZ that greater than UTC+6, because it incorrectly fails to normalize to UTC before extracting the date from a zoned timestamp inside a CRL. The actual CRL has `revocationDate utcTime Time UTCTime 2014-07-07 18:31:35 UTC`.

In UTC+8, that time is not the 7th July; it's the 8th. Need to declare/convert zones properly.


chadlwilson:
One of these tests seems to also be flaky.

chadlwilson:
`test_socket.rb:121` seems flaky to me, don't think I changed anything there.

